PR #19, opened 2017-06-07
Prefer `Data.Text.Lazy` to `Data.Text` in generated code

intractable:
We have client use cases which want to use lazy `Text` values, and it's a better default anyway (e.g., more efficient `(<>)`).


j6carey:
My understanding of the code is that there is no way to change it.  I have been working toward a proposal for adding one, but there are some other things I need to finish first, and it may take some time.

PR #24, opened 2017-08-17
Fix misc parsing/CG bugs

intractable:
The individual commits should be self explanatory, but here's the gist:

- Demote fatal error about `oneof`'s unimplemented state in the code generator to a loud warning; that support is coming soon and it may be useful to be able to process input .protos which use the construct.

- CG: Emit parens around negative integral lits
- CG: Use qualified import for `GHC.Enum` and `GHC.Generics`
- Parser: Resolve ambiguity in `rpcClause` parser
- Parser: Improve comment stripping

### DIFF
--- a/src/Proto3/Suite/DotProto/Generate.hs
+++ b/src/Proto3/Suite/DotProto/Generate.hs
@@ -30,6 +30,7 @@ import           Language.Haskell.Syntax
 import           Language.Haskell.Pretty
 import           System.FilePath
 import           Text.Parsec (ParseError)
+import           Debug.Trace (trace)
 
 -- * Public interface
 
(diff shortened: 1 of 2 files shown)


src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,7 +831,10 @@ apOp :: HsQOp
 apOp  = HsQVarOp (UnQual (HsSymbol "<*>"))
 
 intE :: Integral a => a -> HsExp
-intE = HsLit . HsInt . fromIntegral
+intE x = (if x < 0 then HsParen else id) . HsLit . HsInt . fromIntegral $ x
```
ixmatus: Ah, this is for the signed, negative number case, yah?

intractable: Yessir.

intractable:
@ixmatus Thanks, I agree re: the comment stripping -- it's intended as a stopgap that's a minor improvement on the previous method, which was preventing some `.proto` files from successfully parsing.

PR #25, opened 2017-08-17
`compile-proto-file`: support include paths and output dir, related CG changes

intractable:
#### Overview 

This PR changes the behavior of `compile-proto-file` in significant ways, notably:

- It no longer emits generated code to stdout, but rather to a Haskell module file underneath the directory hierarchy whose root is given by the new `--out` option

- It allows the include search path to be specified via the (repeatable) `--includeDir` option; if the option is not given, the current working directory is used as the sole search path.

During code generation, rather than mapping `.proto` package names to Haskell module names, we now compute a "module path" -- that is, the `typeLikeName`-d version of the `--includeDir`-relative path where the `.proto` file was found. E.g.:
- `foo.proto` maps to a Haskell module named `Foo`
- `foo_bar.proto` maps to a Haskell module named `FooBar`
- `google/protobuf/timestamp.proto` maps to a Haskell module named `Google.Protobuf.Timestamp`

Also, note that `.`characters in the name are treated as `/` characters, so `google.protobuf.timestamp.proto` would yield the same module name and module filename as `google/protobuf/timestamp.proto`.

This is essentially the same module naming scheme as the `protoc` Python plugin uses when compiling `.proto` files.

#### Small example (lifted from the updated documentation also included in this PR)

Let's assume this is our current directory structure before performing any code generation:

```
.
├── my_protos
│   └── my_package.proto
└── other_protos
    └── google
        └── protobuf
            ├── duration.proto
            └── timestamp.proto
```

where `my_package.proto` contains:

```
syntax = "proto3";
package my_package;
import "google/protobuf/timestamp.proto";
import "google/protobuf/duration.proto";
message MyMessage {
  Timestamp timestamp = 1;
  Duration  duration  = 2;
}
```

Then, after the following commands:

```bash
$ compile-proto-file --out gen --includeDir my_protos --includeDir other_protos --proto google/protobuf/duration.proto
$ compile-proto-file --out gen --includeDir my_protos --includeDir other_protos --proto google/protobuf/timestamp.proto
$ compile-proto-file --out gen --includeDir my_protos --includeDir other_protos --proto my_package.proto
```

the directory tree will look like this:

```
.
├── gen
│   ├── Google
│   │   └── Protobuf
│   │       ├── Duration.hs
│   │       └── Timestamp.hs
│   └── MyPackage.hs
├── my_protos
│   └── my_package.proto
└── other_protos
    └── google
        └── protobuf
            ├── duration.proto
            └── timestamp.proto
```

#### Misc

- New or updated dependencies: `doctest`, `foldl`, `neat-interpolation`, `lens`, `optparse-{applicative,generic}`, `system-filepath`, `turtle`

- There is some noise around renaming `test-files/test{_import}.proto`, regenerating code, deleting dead modules such as `tests/OldTestTypes.hs`, etc. -- apologies!


src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -0,0 +1,138 @@
...
+-- >>> toModulePath "relative/path/to/file_without_proto_suffix_fails.txt"
+-- Left "expected .proto suffix"
+--
+-- >>> toModulePath "google/protobuf/timestamp.proto"
+-- Right (Path ["Google","Protobuf","Timestamp"])
+--
+-- >>> toModulePath "a/b/c/google.protobuf.timestamp.proto"
+-- Right (Path ["A","B","C","Google","Protobuf","Timestamp"])
+--
+-- >>> toModulePath "foo/FiLeName_underscore.and.then.some.dots.proto"
```
ixmatus: I really like doctests...

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -0,0 +1,138 @@
...
+-- Right (Path ["Foo","FiLeName_underscore","And","Then","Some","Dots"])
+--
+toModulePath :: FilePath -> Either String Path
+toModulePath fp
+  | Turtle.absolute fp
+    = Left "expected include-relative path"
+  | Turtle.extension fp /= Just "proto"
+    = Left "expected .proto suffix"
+  | otherwise
```
ixmatus: How does this handle `./` and `../`?

ixmatus: `filesystem-filepath` has a `collapse`  function that might be useful, I think.

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -0,0 +1,138 @@
...
+      This is because we currently use the include-relative name to decide
+      the structure of the Haskell module tree that we emit during code
+      generation.
+      |]
+  | otherwise = case toModulePath protoFP of
+      Left e -> pure (BadModulePath e)
+      Right mp -> do
+        mfp <- flip Turtle.fold FL.head $ do
+          fp <- (</> protoFP) <$> Turtle.select searchPaths
+          True <- Turtle.testfile fp
```
ixmatus: Minor suggestion. You could move the `</> protoFP` to here:

```
True <- Turtle.testfile (fp </> protoFP)
```

Which simplifies the `Turtle.select`, to me. This is non-critical feedback, though.

src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -19,12 +22,25 @@ import Text.Parser.Char
...
-
-parseProto :: String -> Either ParseError DotProto
-parseProto = parse topLevel "" . stripComments
+-- interfaces
+
+-- | @parseProto mp inp@ attempts to parse @inp@ as a 'DotProto'. @mp@ is the
+-- module path to be injected into the AST as part of 'DotProtoMeta' metadata on
+-- a successful parse.
+parseProto :: Path -> String -> Either ParseError DotProto
+parseProto modulePath = parse (topLevel modulePath) "" . stripComments
```
ixmatus: What is the reason for feeding the parsec parser a `String` instead of a `Text`? (other than `stripComments` returning a `String`...)

src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -19,12 +22,25 @@ import Text.Parser.Char
...
+
+-- | @parseProtoFile mp fp@ reads and parses the .proto file found at @fp@. @mp@
+-- is used downstream during code generation when we need to generate names
+-- which are a function of the source .proto file's filename and its path
+-- relative to some @--includeDir@.
+parseProtoFile :: Turtle.MonadIO m
+               => Path -> Turtle.FilePath -> m (Either ParseError DotProto)
+parseProtoFile modulePath =
+  fmap (parseProto modulePath) . Turtle.liftIO . readFile . FP.encodeString
```
ixmatus: If we fed the parser a `Text` stream this could be simplified to:

```haskell
fmap (parseProto modulePath) . Turtle.readTextFile
```

src/Proto3/Suite/DotProto/Internal.hs
intractable: Great catch, thanks.

src/Proto3/Suite/DotProto/Parsing.hs
intractable: Yeah, it really shouldn't be `String`, but there's a comment:

```-- | This module contains a near-direct translation of the proto3 grammar
--   It uses String for easier compatibility with DotProto.Generator, which needs it for not very good reasons
```

It's on the list of things to improve, but (a) we're likely to obsolete this parser and (b) I don't want to couple that change with this PR, even if it's likely to be easy.

intractable:
@ixmatus Yep, thanks, and thanks for catching the bug. 

Let me know if there are any other outsanding issues before I merge.

Your'e right that the test file noise is from rename refactors since the `.proto` filenames changed and such.

src/Proto3/Suite/DotProto/Parsing.hs
ixmatus: Oh okay. Yeah that makes sense.

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -57,6 +58,9 @@ dieLines (Turtle.textToLines -> msg) = do
 -- >>> toModulePath "foo/FiLeName_underscore.and.then.some.dots.proto"
 -- Right (Path ["Foo","FiLeName_underscore","And","Then","Some","Dots"])
 --
+-- >>> toModulePath "foo/bar/././baz/../boggle.proto"
```
ixmatus: I think there's still a problem if the path begins with `./` or `../`. I don't think `collapse` will do anything about those. From documentation, I think `collapse "./foo/bar" == "./foo/bar"` so you may have to simply drop `./` from the beginning if it's there.

For the case in which the path starts with `../`, you may want that to simply be an error.

intractable:
Looks like I need to handle a couple of more edge cases in `toModulePath`, will do that shortly.

intractable:
@ixmatus Take another peek and let me know what you think? The edge cases now covered are:

- Preceding `./`: strip it
- Preceding `.`: strip it and ignore it (`foo.proto` same as`.foo.proto`)
- Preceding `../`: raise an error, as it's no longer "under" a search path
- `..` substring after `collapse`: error; this rejects stuff like `foo..proto` outright since we use the separators to infer directory structure

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -61,22 +75,37 @@ dieLines (Turtle.textToLines -> msg) = do
...
 --
+-- >>> toModulePath "./foo.proto"
+-- Right (Path ["Foo"])
+--
+-- NB: We ignore preceding single '.' characters
+-- >>> toModulePath ".foo.proto"
+-- Right (Path ["Foo"])
 toModulePath :: FilePath -> Either String Path
-toModulePath fp
+toModulePath fp0@(fromMaybe fp0 . FP.stripPrefix "./" -> fp)
```
ixmatus: This is slick. I didn't realize the binding from `@` would be in scope for a view pattern though it makes sense when I think about it (I guess it expands that to an openly recursive let binding?)

intractable:
Doing some integration testing prior to merge, should be good to go relatively soon.

PR #26, opened 2017-08-23
Derive Enum for protobuf enumerations

themattchan:


themattchan:
Nvm, enums are generated somewhere else...

PR #29, opened 2017-09-07
Preliminary JSON encodings of protobuf payloads

intractable:
#### Overview

This PR provides partial support for the [proto3 "jsonpb" JSON encoding scheme](https://developers.google.com/protocol-buffers/docs/proto3#json) and associated code generation.

The code submitted in this PR:

- Defines two typeclasses, `FromJSONPB` and `ToJSONPB`; they resemble their `aeson` counterparts, with a similar surface syntax provided for familiar-looking instance declarations

- Generates the appropriate instances of the above typeclasses, for message and enum types (see examples below)

- Provides the "jsonpb" encoding for all scalar types and the most common aggregate types
  - NB: It does not currently support `map<K,V>`, `Any`, or the special named types described in the type table above. 
  - Providing "jsonpb" support for e.g. `Duration` and `Timestamp` should probably be next on the list to support.

- Allows explicit emission of default values in the JSON encoding; defaults to `False`, controllable via the appropriate `Options` type.

- Now depends on `aeson-1.1.1.0`

- Has been only lightly tested via some round-trip doctest props 

#### Examples

1. A protobuf message 
```
message OutOfOrderFields {
  repeated uint32 field1 = 2001 [packed = true];
  string          field2 = 101;
  int64           field3 = 30;
  repeated string field4 = 1002;
}
```
now entails generation of instances like
```
instance HsJSONPB.ToJSONPB OutOfOrderFields where
        toEncodingPB opts (OutOfOrderFields f2001 f101 f30 f1002)
          = (HsJSONPB.fieldsPB opts
               ["field1" .= f2001, "field2" .= f101, "field3" .= f30,
                "field4" .= f1002])
 
instance HsJSONPB.FromJSONPB OutOfOrderFields where
        parseJSONPB
          = (HsJSONPB.withObject "OutOfOrderFields"
               (\ obj ->
                  (Hs.pure OutOfOrderFields) <*> obj .: "field1" <*> obj .: "field2"
                    <*> obj .: "field3"
                    <*> obj .: "field4"))
```

2. A protobuf enum
```  
enum TestEnum {
    ENUM1 = 0;
    ENUM2 = 1;
    ENUM3 = 2;
  }
```
now entails generation of instances like
```
instance HsJSONPB.ToJSONPB TestEnum where
        toEncodingPB _ = HsJSONPB.namedEncoding
 
instance HsJSONPB.FromJSONPB TestEnum where
        parseJSONPB (HsJSONPB.String "ENUM1")
          = Hs.pure TestEnumENUM1
        parseJSONPB (HsJSONPB.String "ENUM2")
          = Hs.pure TestEnumENUM2
        parseJSONPB (HsJSONPB.String "ENUM3")
          = Hs.pure TestEnumENUM3
        parseJSONPB v = (HsJSONPB.typeMismatch "TestEnum" v)
```


src/Proto3/Suite/Class.hs
```diff
@@ -197,6 +197,32 @@ instance HasDefault (Nested a) where
...
+  isDefault = isNothing
+
+-- TODO: Determine if we have a reason for rendering fixed32/sfixed as Fixed
+-- Word32/Int32 in generated datatypes; for other field types, we omit the
+-- newtype wrappers in the type signature but un/wrap them as needed in the
+-- encode/decodeMessage implementations. These Fixed wrappers can probably be
+-- removed and the type interface would be more consistent with other types, but
+-- until that occurs, the following two instances are needed.
+--
+-- Tracked by https://jira.mv.awakenetworks.net/browse/MONAPP-7986
```
ixmatus: Perhaps duplicate that ticket to a github issue here as well?

intractable: Oops, thanks.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -546,6 +555,67 @@ messageInstD ctxt parentIdent msgIdent messageParts =
...
+  pure (instDecl_ (jsonpbName "ToJSONPB")
+                 [ type_ msgName ]
+                 [ HsFunBind [ toEncodingPBDecl ] ])
+
+fromJSONPBMessageInstD :: TypeContext -> DotProtoIdentifier -> DotProtoIdentifier
+                       -> [DotProtoMessagePart] -> CompileResult HsDecl
+fromJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts =
+  do
+  msgName <- nestedTypeName parentIdent =<< dpIdentUnqualName msgIdent
+  kvps    <- sequence
```
ixmatus: Since this same code appears in `toJSONPBMessageInstD` I think factoring it out into a separate function is a good idea. This is a nit though so it's non-critical.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -546,6 +555,67 @@ messageInstD ctxt parentIdent msgIdent messageParts =
                       , decodeMessageDecl
                       , dotProtoDecl ]))
 
+toJSONPBMessageInstD :: TypeContext -> DotProtoIdentifier -> DotProtoIdentifier
+                     -> [DotProtoMessagePart] -> CompileResult HsDecl
+toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts =
```
ixmatus: I'm not sure how you feel about this so it's up to your judgement, but I only _know_ what the generated code mostly looks like from reading the PR description. Which leads me to think a comment providing a very limited example of an out-of-order field number-tagged to instance will be helpful for posterity and maintenance.

(same comment for `fromJSONPBMessageInstD`.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -685,12 +755,45 @@ dotProtoEnumD parentIdent enumIdent enumParts =
          predFailure     = match_ (HsIdent "pred") [ HsPWildCard ]
                                   (HsUnGuardedRhs (HsApp predErrorE enumNameE)) []
 
+         parseJSONPBDecls :: [HsMatch]
+         parseJSONPBDecls =
+           [ let pat nm =
```
ixmatus: Large list comprehensions like this one, while not _difficult_ to parse, certainly take a minute to wrap my head around. Perhaps do notation in the list monad would make this more readable?

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+
+-- * JSONPB codec entry points
+
+-- | 'Data.Aeson.encode' variant for serializing a JSONPB value as a lazy
+-- 'LBS.ByteString'.
+encode :: ToJSONPB a => Options -> a -> LBS.ByteString
+encode opts = E.encodingToLazyByteString . toEncodingPB opts
+{-# INLINE encode #-}
+
+-- | 'Data.Aeson..eitherDecode' variant for deserializing a JSONPB value from a
```
ixmatus: Nit, extra `.` in `Data.Aeson..eitherDecode`.

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+  where
+    (.:?) = A.explicitParseFieldMaybe parseJSONPB
+
+-- * JSONPB rendering and parsing options
+
+data Options = Options
+  { optEmitDefaultValuedFields :: Bool
+  }
+
+defaultOptions :: Options
```
ixmatus: I've found that it can be really nice (not necessary though, this is a nit) in generated documentation to see what the default values are.

It creates a bigger maintenance burden and is likely not worth doing since `Options` here is such a small record, but I have found myself appreciating `defaultOptions`-style producers that document what the default values are without forcing me to go look at the source (I know, I'm lazy).

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+defaultOptions = Options
+  { optEmitDefaultValuedFields = False
+  }
+
+-- * Helper types and functions
+
+dropNamedPfx :: Named a => Proxy a -> String -> String
```
Gabriel439: Minor suggestion: rename to `dropNamedPrefix`

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+-- object, or if it is present but its value is null, we produce the default
+-- protobuf value for the field type
+(.:) :: (FromJSONPB a, HasDefault a) => A.Object -> Text -> A.Parser a
+obj .: key = obj .:? key A..!= def
+  where
+    (.:?) = A.explicitParseFieldMaybe parseJSONPB
+
+-- * JSONPB rendering and parsing options
+
+data Options = Options
```
ixmatus: In-fact, you could add a show instance to this. Then add a simple doc test on `defaultOptions` which would accomplish showing what the default values are. Not sure if that's necessary though.

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+  { fieldIsDefaultValued :: Bool
+    -- ^ Whether or not this field is "default-valued"; i.e., whether the value
+    -- component of the key-value pair is the protobuf default value for its
+    -- type
+  , fieldTuple           :: Options -> A.Series
+    -- ^ Produce the JSONPB-encoded key-value tuple
+  }
+
+-- | @fieldsPB opts flds@ encodes the given JSONPB-encoded fields
+fieldsPB :: Options -> [FieldPB] -> A.Encoding
```
ixmatus: Subjective opinion nit (feel free to ignore entirely): if this function encodes JSONPB encoded fields then why not name it `encodePBFields` or `encodeFields`? I also think the doc comment for this function is not carrying its own weight.

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+fieldsPB opts@Options{..} fields = E.pairs (mconcat (fmap emit fields))
+  where
+    emit FieldPB{..}
+      | not optEmitDefaultValuedFields && fieldIsDefaultValued
+        = mempty
+      | otherwise
+        = fieldTuple opts
+
+namedEncoding :: forall e. (Named e, Show e) => e -> A.Encoding
+namedEncoding = E.string . dropNamedPfx (Proxy :: Proxy e) . show
```
ixmatus: Would `TypeApplications` be useful instead of `Proxy`? I can't remember if you explored that or not.

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -0,0 +1,193 @@
...
+      where
+        skipSpace :: Atto.Parser ()
+        skipSpace = Atto.skipWhile $ \w -> w == 0x20 || w == 0x0a || w == 0x0d || w == 0x09
+        {-# INLINE skipSpace #-}
+{-# INLINE eitherDecode #-}
+
+-- * Operator definitions
+
+-- | Construct a (field name, JSONPB-encoded value) tuple
+(.=) :: (HasDefault v, ToJSONPB v) => Text -> v -> FieldPB
```
Gabriel439: Perhaps this could instead have type:

```haskell
(.=) :: (HasDefault v, ToJSONPB v) => Text -> v -> Option -> A.Encoding
(.=) = ... -- It uses `emit` internally to directly produce the desired encoding
```

The reason I suggest this is then you could simplify the `toEncodingPB` example to:

```haskell
toEncodingPB opts (Scalar32 i32 u32 s32 f32 sf32) = mconcat
    [ "i32"  .= i32
    , "u32"  .= u32
    , "s32"  .= s32
    , "f32"  .= f32
    , "sf32" .= sf32
    ] opts
```

... and if you changed the order of argument for `toEncodingPB` you could simplify it further to:

```haskell
toEncodingPB (Scalar32 i32 u32 s32 f32 sf32) = mconcat
    [ "i32"  .= i32
    , "u32"  .= u32
    , "s32"  .= s32
    , "f32"  .= f32
    , "sf32" .= sf32
    ]
```

Gabriel439:
Sorry for the duplicate review comments.  GitHub was misbehaving

src/Proto3/Suite/JSONPB/Class.hs
Gabriel439: Yeah, this works if you use `TypeApplications` and `AllowAmbiguousTypes`

src/Proto3/Suite/JSONPB/Instances.hs
```diff
@@ -0,0 +1,183 @@
...
+
+-- sfixed32
+instance ToJSONPB (Fixed Int32) where
+  toEncodingPB opts = toEncodingPB opts . fixed
+instance FromJSONPB (Fixed Int32) where
+  parseJSONPB = fmap Fixed . parseJSONPB
+
+-- sfixed64
+instance ToJSONPB (Fixed Int64) where
+  toEncodingPB opts = toEncodingPB opts . fixed
```
Gabriel439: For all of these instances that just lift `toEncodingPB` over a newtype wrapper, you can use:

```haskell
toEncodingPB = coerce (toEncodingPB @Int64)
```

ixmatus:
@intractable I didn't say it before but this looks really good BTW.

src/Proto3/Suite/DotProto/Generate.hs
intractable: I think it's only marginally more readable via use of the list monad, but then it sticks out stylistically from the rest of the module which makes regular use of list comprehensions. /shrug. Going to leave it as-is, but point taken.

src/Proto3/Suite/JSONPB/Class.hs
intractable: I was thinking that `A.pairs : A.Pair :: fieldsPB : FieldPB`...agreed on the comment.

src/Proto3/Suite/JSONPB/Class.hs
intractable: Good idea. I always forget about the `Monoid` instance for functions. 

It's slightly more involved than what you suggested, since `A.Encoding` is not a `Monoid` and we need to apply `Data.Aeson.Encoding.pairs`, but I've still simplified considerably it by swapping the argument order and getting rid of the intermediate `FieldPB` type.

PR #31, opened 2017-09-12
Tighten upper bound on QuickCheck

kim:
QuickCheck 2.10 moves some instances to the `quickcheck-instances` package, notably for `Natural`.

This is mostly a heads-up, I'd also be happy to provide a CPP'd patch if you'd prefer that.


Gabriel439:
Thanks!  I just independently confirmed that we need to add `quickcheck-instances` before we can increase the upper bound on `QuickCheck`

PR #32, opened 2017-09-12
Fix `CanonicalRank` default signature for GHC 8.2

kim:
GHC 8.2 became more restrictive when checking default signatures, demanding an
explicit equality constraint here. See also:

https://github.com/ekmett/lens/issues/712

Referenced there:
https://ghc.haskell.org/trac/ghc/ticket/13258
https://ghc.haskell.org/trac/ghc/ticket/13249


Gabriel439:
Thanks for the fix!

PR #40, opened 2017-11-08
Add support for packed repeated bools

Gabriel439:
The protocol buffers specification says that:

> Only repeated fields of primitive numeric types (types which use the varint,
> 32-bit, or 64-bit wire types) can be declared "packed".

However, bool values fall under this "numeric" category because they are
technically encoded as varints.  In Haskell terms, this means that there needs
to be a `MessageField (PackedVec Bool)` instance.

This change adds the new instance and also updates the test suite to verify that
the added instance interops with the Python implementation.  The Python
implementation also provides evidence that a packed repeated bool is supported
by at least one other implementation.


intractable:
LGTM; I'm amidst making a similar change for packed vecs of enums as well.

PR #42, opened 2018-01-12
Make instance generation extensible

gbaz:
This PR lets the command line utility take a flag that points to an additional haskell file, which needs to be syntactically correct, but need not compile on its own.

Imports from this file will be added to the generated HS, and instances from this file will be used to replace instances that would otherwise have been generated.

This lets users override json instance creation, for example, to sub in more friendly serialization formats...


intractable:
This will be very useful, thanks! 

The only thing that makes me a little hesitant to merge is the handling of namespacing and the lack of a test case. It'd also be nice to have some documentation for a trivial use of the feature, but a clear test case could stand in for that if needed.

Re the namespacing, maybe the CLI interface ought to allow specification of a fully qualified (`package`-scoped) name to the filename of the module which provides the instance override. What do you think?

gbaz:
Ok, i fixed the namespacing. I don't have rights to merge -- can someone else?

PR #43, opened 2018-01-18
Parse options for enum types

themattchan:
Currently only parses the options but does not use them in the generated code.

Additionally, this PR fixes type errors in the test suite introduced by #42


themattchan:
@intractable do we want to generate `DEPRECATED` comments from the `deprecated` option? (And if so, should I merge this now and make a separate PR?)

intractable:
@themattchan It'd be nice to do that, yes, but a follow-on PR is fine.

PR #44, opened 2018-01-18
Fix canonicalize-proto-file

themattchan:
Fixes the regression introduced by #43


Gabriel439:
Thank you for fixing this!

PR #48, opened 2018-01-19
Update nixpkgs to release-17.09

themattchan:
Fixes #47

PR #49, opened 2018-01-19
Enable Travis CI

themattchan:
Build status for this PR here: https://travis-ci.org/awakesecurity/proto3-suite/

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,2 @@
+language: nix
+script: nix-channel --remove nixpkgs && nix-channel --add https://nixos.org/channels/nixos-16.09/ nixpkgs && nix-channel --update && nix-build release.nix
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,2 +1,2 @@
 language: nix
-script: nix-channel --remove nixpkgs && nix-channel --add https://nixos.org/channels/nixos-16.09/ nixpkgs && nix-channel --update && nix-build release.nix
+script: nix-build --attr proto3-suite-linux release.nix
--- a/README.md
+++ b/README.md
@@ -1,5 +1,7 @@
 # `proto3-suite`
 
+[![Build Status](https://travis-ci.org/awakesecurity/proto3-suite.svg?branch=master)](https://travis-ci.org/awakesecurity/proto3-suite)
+
 This package defines tools for working with protocol buffers version 3 in Haskell.
 
 This library provides a higher-level API to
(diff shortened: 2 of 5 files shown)


.travis.yml
```diff
@@ -0,0 +1,2 @@
+language: nix
+script: nix-channel --remove nixpkgs && nix-channel --add https://nixos.org/channels/nixos-16.09/ nixpkgs && nix-channel --update && nix-build release.nix
```
Gabriel439: We don't need this anymore.  This was for before we learned how to pin `nixpkgs` without the `NIX_PATH`.  See: https://nixos.wiki/wiki/How_to_fetch_Nixpkgs_with_an_empty_NIX_PATH

If you use the above recipe then the only command you need here is `nix-build release.nix`

themattchan:
Seems like the build is failing because the CI server is trying to build the `darwin` version on a `linux` box. @Gabriel439 is there an option to make nix only build `proto3-suite-linux`?

.travis.yml
themattchan: How do you integrate the script with the `nixpkgs.json` parsing here? https://github.com/awakesecurity/proto3-suite/blob/master/release.nix#L88

Maybe something like this?
``` 
 nixpkgs =
    let
      json = builtins.fromJSON (builtins.readFile ./nixpkgs.json);
    in fetchNixpkgs { rev = json.rev; sha = json.sha; };
```

Gabriel439: That won't work because the `sha256` that `fetchNixpkgs` expects is computed in a different way.  `gRPC-haskell` has a good example of instructions on how to do this: https://github.com/awakesecurity/gRPC-haskell/blob/master/nixpkgs.nix

Gabriel439:
@themattchan: Yes.  Do: `nix-build --attr proto3-suite-linux release.nix`

Gabriel439:
There's one more thing I would suggest, which is to set up Travis CI to only run on pull requests instead of on every branch.  That helps avoid wasteful runs when people are doing development work.  There is an option in the project settings page for this.

themattchan:
Good idea -- I've updated the settings for `proto3-wire` as well.

themattchan:
Right now this is still on `nixpkgs 16.09`, should I merge this PR with #48?

Gabriel439:
@themattchan: Yeah, that's fine

themattchan:
On second thought, let's get this into master first while I debug the nixpkgs upgrade.

PR #52, opened 2018-01-31
Add support for incremental parsing.

j6carey:
NOTE: Must wait for [proto3-wire PR 35](https://github.com/awakesecurity/proto3-wire/pull/35).


j6carey:
Actually, never mind; the application that I had in mind for this won't work, because a top-level protobuf message is not self-delimited, so you can't tell when it ends.

PR #54, opened 2018-03-24
Replace `CompileResult` with `MonadError CompileError`

Gabriel439:
This updates the code-generation logic to use `MonadError` so that the same
utilities can be reused between `Either` and `ExceptT` without conversion


Gabriel439:
Note that CI is failing due to using Nix 2.0, but I verified that this builds locally.

We'll have to eventually fix the `nixpkgs` derivation to be compatible with Nix 1.* and Nix 2.*

PR #57, opened 2018-03-29
Swagger-compatible JSON encoding

themattchan:
**BREAKING CHANGES**

If you previously relied on the `HsJSONPB.ToJSON` methods to generate JSONPB compatible json, you now need to use the `HsJSONPB.ToJSONPB` methods, manually supplying `jsonpbOptions`.
Also, `defaultOptions` are no longer the JSONPB defaults; that has been renamed as the aforementioned `jsonpbOptions`.

---
This PR fixes a compatibility issue between `swagger2` and JSONPB's encoding of
`oneof` message fields, by changing the default behaviour of the JSONPB instance
to generate non-inlined serialisations. The motivation behind this is that our
internal Swagger API schemas generated from the generated Haskell data types
generate *named* sum type fields, breaking the API for consumers of the
swagger schema.

This is controlled by the option supplied to the `ToJSONPB` methods, `optEmitInlinedOneof`. 
The default is set to `false` (this breaks compatibility with the Go library but is a saner encoding).
The `FromJSONPB` instance accepts both encodings.

For example, given the following protobuf definition

```proto
message Foo {
  string foo = 1;
}

message Bar {
  int32 bar = 1;
}

message MyMessage {
  oneof FooOrBar {
    Foo x = 1;
    Bar y = 2;
  }  
}
```

and a message to encode:

```haskell
-- Generated code
data Foo = Foo{fooFoo :: Hs.Text}
data Bar = Bar{barBar :: Hs.Int32}
data MyMessage = MyMessage{myMessageFooOrBar ::
                           Hs.Maybe MyMessageFooOrBar}
data MyMessageFooOrBar = MyMessageFooOrBarX Example.Foo
                       | MyMessageFooOrBarY Example.Bar

message = MyMessage (MyMessageFooOrBarX (Foo "hello"))
```

the go-jsonpb compliant encoding (the previous default) for the following message is

```json
{ "x": "hello" }
```

the new default will be this

```json
{ "MyMessageFooOrBar": { "x": "hello" } }
```


src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -103,11 +105,22 @@ class ToJSONPB a where
   toEncodingPB :: a -> Options -> A.Encoding
   toEncodingPB x = A.toEncoding . toJSONPB x
 
+instance ToJSONPB A.Value where
+  toJSONPB v _ = v
+  toEncodingPB v _ = E.value v
+
+instance ToJSONPB A.Encoding where
+  toJSONPB e _ = fromJust . A.decode . Builder.toLazyByteString . E.fromEncoding $ e
```
ixmatus: `fromJust` is not safe and will throw an exception if `A.decode` produces a `Nothing`. What is the rationale behind this behavior and can we explain that rationale in a comment?

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -880,40 +943,58 @@ fromJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
   msgName    <- nestedTypeName parentIdent =<< dpIdentUnqualName msgIdent
   qualFields <- getQualifiedFields msgName messageParts
 
+  let lambdaPVar = patVar "obj"
+  let lambdaVar  = HsVar (unqual_ "obj")
   -- E.g., for message Something{ oneof name_or_id { string name = _; int32 someid = _; } }:
```
ixmatus: Is this example still accurate?

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,48 +834,103 @@ toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
+  -- >   <*> pure optEmitInlinedOneof
+  --
+  -- ==>
+  --
+  -- > (liftA3 (liftA3 bool))
+  -- >   ((flip ("Foo" .=) <*>) . PB.object . pure)
+  -- >   id
+  -- >   (pure optEmitInlinedOneof)
+  --
+  -- The interested reader is invited to verify that this transformation is
```
ixmatus: I know you don't mean it this way but I read this with a patronizing voice. I know this is standard fare in textbooks but I think we can do better in our opensource code by providing a test to show that it is a correct transformation (if you haven't already); thus, also enlightening the reader.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,48 +834,103 @@ toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
+  -- >   HsJSONPB.object
+  -- >     [ (\caseExpr option -> if optEmitInlinedOneof option
+  -- >                            then oneofCaseE option
+  -- >                            else PB.object ["Foo" .= (PB.object [caseExpr] option)] option
+  -- >       ) (<case expr scrutinising foo> :: Options -> Value)
+  -- >     , <do more>
+  -- >     , <do stuff>
+  -- >     ]
+  --
+  -- To avoid gensymming parameters for the lambda, we will instead generate the
```
ixmatus: What does "gensymming parameters" mean?

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,48 +834,103 @@ toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
-  -- case f4_or_f9 of
-  --   Just (SomethingPickOneName f4)
-  --     -> HsJSONPB.pair "name" f4
-  --   Just (SomethingPickOneSomeid f9)
-  --     -> HsJSONPB.pair "someid" f9
-  --   Nothing
-  --     -> mempty
-  let oneofCaseE (OneofField{subfields}) =
-        HsCase disjunctName (altEs <> pure fallThruE)
+  -- Note: [Generating eta-reduced code]
```
ixmatus: What is the benefit of generating eta-reduced code?

src/Proto3/Suite/Class.hs
```diff
@@ -583,7 +583,9 @@ instance (MessageField e, KnownSymbol comments) => MessageField (e // comments)
 decodePacked
   :: Parser RawPrimitive [a]
   -> Parser RawField (PackedVec a)
-decodePacked p = Parser $ \fs -> fmap (fromList . join . F.toList) $ TR.sequence $ fmap (runParser p) fs
+decodePacked = Parser
```
ixmatus: I think some of your other point-free and `fmap` -> `<$>` changes in this file are reasonable and maintain readability (if not increase it in certain cases); however, I don't think that is the case with the change on this line. This required more attention from me to penetrate its meaning.

src/Proto3/Suite/DotProto/Generate.hs
ixmatus: Joel explained to me that "gensymming" is "generating symbols for". IIRC from his older PRs this was intentional so that generated code would be easier to understand by visual inspection alongside the input protos.

src/Proto3/Suite/DotProto/Generate.hs
ixmatus: I also think that if there aren't any clear benefits to the eta-reduced code, then code that _does not require_ the reader to go through the exercise of program calculation (or following a test verifying the transformation) is better and clearer.

src/Proto3/Suite/DotProto/Generate.hs
ixmatus: For instance, I think @intractable gave good rationale for this in the description of this PR: https://github.com/awakesecurity/proto3-suite/pull/29

Gabriel439: I don't think we should use the reader monad just to avoid generating unique names.  This makes the generated code hard for me to read.  For example `((flip ("Foo" .=) <*>) . PB.object . pure)` is difficult for me to read

src/Proto3/Suite/Class.hs
Gabriel439: This was also difficult for me to read.  Actually, even the original code before your change was difficult for me to read, too :)

I did like the simplification of `sequence . fmap f → traverse f`, but overall I still think this is too point-free

Also, independent of point-free versus point-ful, I think another way to improve the clarify of this code is to split the `fromList . join . F.toList` out into a named function, like this:

```haskell
decodePacked = ...
  where
    adapt :: Seq [a] -> Vector a
    adapt = fromList . join . F.toList
```

src/Proto3/Suite/Class.hs
```diff
@@ -605,7 +607,7 @@ class Message a where
 
 -- | Generate metadata for a message type.
 message :: (Message a, Named a) => Proxy a -> DotProtoDefinition
-message pr = DotProtoMessage (Single $ nameOf pr) $ DotProtoMessageField <$> (dotProto pr)
+message = DotProtoMessage <$> (Single . nameOf) <*> (fmap DotProtoMessageField . dotProto)
```
Gabriel439: I think the original version was more clear.  My usual rule of thumb is that if a piece of code uses `Functor`/`Applicative`/`Monad` in more than one way then I usually monomorphize all but one of the uses.  For example, here you use `<$>`/`<*>` in one case for `Reader (Proxy a)` and `fmap` in another way for `[]`

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -713,7 +717,7 @@ messageInstD ctxt parentIdent msgIdent messageParts =
             FieldNormal _fieldName fieldNum dpType options ->
                 do fieldE <- wrapE ctxt dpType options recordFieldName'
                    pure (apply encodeMessageFieldE [ fieldNumberE fieldNum, fieldE ])
-            FieldOneOf OneofField{subfields} ->
+            FieldOneOf (OneofField _ subfields) ->
```
Gabriel439: I prefer the original version because it works regardless of how many fields the `OneofField` constructor has, resulting in less code churn as people extend the data type

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -167,16 +181,27 @@ parseField :: FromJSONPB a
...
+instance HasDefault A.Value where
+  def       = A.Null
+  isDefault = (== A.Null)
+
 -- * JSONPB rendering and parsing options
 
 data Options = Options
   { optEmitDefaultValuedFields :: Bool
+  , optEmitInlinedOneof :: Bool
+  -- ^ for compatibility with Go JSONPB.
```
Gabriel439: There should be a much longer comment somewhere (possibly here, but could be somewhere else) explaining what is the issue with the Go JSONPB encoding.  Specifically, it should clarify that the JSONPB way of representing `oneOf` is incompatible with the OpenAPI v3.0 way of representing `oneOf`

src/Proto3/Suite/DotProto/Generate.hs
Gabriel439: I want to further stress the importance of making the generated code easier to read by quoting this:

> Everyone knows that debugging is twice as hard as writing a program in the first place. So if you're as clever as you can be when you write it, how will you ever debug it?
>
> - Brian Kernighan

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -769,11 +773,13 @@ messageInstD ctxt parentIdent msgIdent messageParts =
         | QualifiedField _ fieldInfo <- qualifiedFields ]
 
      dotProtoE <- HsList <$> sequence
-         [ dpTypeE dpType >>= \typeE ->
-             pure (apply dotProtoFieldC [ fieldNumberE fieldNum, typeE
-                                        , dpIdentE fieldIdent
-                                        , HsList (map optionE options)
-                                        , maybeE (HsLit . HsString) comments ])
+         [ dpTypeE dpType <&> \typeE ->
```
Gabriel439: Generous use of `let` bindings can mitigate dependence on more exotic operators.  For example:

```haskell
-- I like to use `adapt` when I'm not sure what to name a function, sort of like how
-- people use `go` for loops
[ let adapt typeE = apply dotProtoFieldC ...
  in  fmap adapt (dpTypeE dpType)
, ...
```

src/Proto3/Suite/Class.hs
themattchan: Yeah, I just did this via rewrite rules, I'm not sure what's going on either.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,48 +834,103 @@ toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
-  let oneofCaseE (OneofField{subfields}) =
-        HsCase disjunctName (altEs <> pure fallThruE)
+  -- Note: [Generating eta-reduced code]
+  -- ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+  -- Suppose we have a sum type Foo, nested inside a message Bar.
+  -- We want to generate the following:
+  --
+  -- > toJSONPB (Bar foo more stuff) =
+  -- >   HsJSONPB.object
+  -- >     [ (\caseExpr option -> if optEmitInlinedOneof option
```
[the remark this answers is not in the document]
themattchan: It can be, but writing it in the desugared form makes the calculation easier

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -826,48 +834,103 @@ toJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
+  -- >                            then oneofCaseE option
+  -- >                            else PB.object ["Foo" .= (PB.object [caseExpr] option)] option
```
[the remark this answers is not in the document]
themattchan: It's `HsJSONPB.object`. 

I got the type of `caseExpr` wrong, though. It's some  `Monoid m` returned by `pair :: ToJSONPB v => Text -> v -> Options -> m`.

The type is `object :: [Options -> [A.Pair]] -> Options -> A.Value`, so it works out like this:

```
caseExpr :: Options -> m

(PB.object [caseExpr] option) :: A.Value

"Foo" .= (PB.object [caseExpr] option) :: KeyValuePB kvp => Options -> kvp

PB.object ["Foo" .= (PB.object [caseExpr] option)] option :: A.Value
```

This is probably why I had to add a few typeclass instances, though.

src/Proto3/Suite/DotProto/Generate.hs
themattchan: This generated code is now nested within more code.

src/Proto3/Suite/Class.hs
themattchan: Factored out a local function.

src/Proto3/Suite/DotProto/Generate.hs
themattchan: I think I did this to fix some error, but it's unnecessary. Reverted

src/Proto3/Suite/Class.hs
ixmatus: I like the pointful version you just suggested the most. That is easy for me to understand at first-glance.

src/Proto3/Suite/DotProto/Generate.hs
themattchan: Moved comment.

src/Proto3/Suite/DotProto/Generate.hs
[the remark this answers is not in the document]
Gabriel439: What calculation do you mean?

src/Proto3/Suite/DotProto/Generate.hs
Gabriel439: I still believe the generated code should be easy for people to read.  Also, all free variables within the example are statically specified by the code generator so I believe there would be no risk of a free variable being shadowed by our bound variable named `options`

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -880,40 +943,60 @@ fromJSONPBMessageInstD _ctxt parentIdent msgIdent messageParts = do
...
+                                    bindOp
+                                    (apply (HsVar (jsonpbName "withObject")) [ oneofTyLit , bndName ])
+
+              parseUnwrapped = HsParen (HsApp bndName lambdaVar)
+
+          -- E.g., for message Something{ oneof name_or_id { string name = _; int32 someid = _; } }:
+          -- parseNameOrId =
+          --   fmap Hs.msum
+          --     (Hs.sequence
+          --        [ fmap (Just . SomethingPickOneName) <$> (`HsJSONPB.parseField` "name")
```
Gabriel439: Same comment here: two levels of `fmap` makes this hard for me to follow what the generated code does

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -167,16 +181,42 @@ parseField :: FromJSONPB a
...
+  isDefault = (== A.Null)
+
 -- * JSONPB rendering and parsing options
 
 data Options = Options
   { optEmitDefaultValuedFields :: Bool
+  , optEmitInlinedOneof :: Bool
+  -- ^ For compatibility with Go JSONPB.
+  --
+  -- If 'True', the following message will be serialized as:
```
Gabriel439: Minor nitpick: move the `... will be serialized as:` in between the protobuf message definition and the first example serialization

src/Proto3/Suite/JSONPB/Class.hs
```diff
@@ -167,16 +181,42 @@ parseField :: FromJSONPB a
...
+  -- instead of
+  --
+  -- > MyMessage (Animal (Cat "Simba")) => { "animal": { "cat": "Simba" } }
   } deriving Show
 
 -- | Default options for JSONPB encoding. By default, all options are @False@.
 defaultOptions :: Options
 defaultOptions = Options
   { optEmitDefaultValuedFields = False
+  , optEmitInlinedOneof = False
```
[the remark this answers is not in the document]
themattchan: Slightly less confusing options would be

```
Options
  { optEmitDefaultValuedFields :: Bool 
  -- ^ false if jsonpb
  , optEmitNamedOneof :: Bool
  -- ^ false if jsonpb
  }
```

src/Proto3/Suite/JSONPB/Class.hs
themattchan: Ok

src/Proto3/Suite/DotProto/Generate.hs
ixmatus: I think explicitness increases clarity in most cases by reducing indirection. There are cases where pointfree code is more natural to read and it makes sense; however, I think a lot of pointfree code introduces too much indirection, reducing clarity. 

I think clarity is important because it lowers the "gumption barrier" for contributors familiar and not familiar with the codebase. Clarity is also important for users who may have to debug issues or figure out how to use the generated code.

The fact that we have complete control over what is generated only reinforces that we should do the most clear thing for our users and for contributors.

src/Proto3/Suite/DotProto/Generate.hs
themattchan: Ok, I'll add the parameter back in, at the cost of some generated binder verbosity

```
parseNameOrId parseNameOrId_object = 
  Hs.msum
    [ Just . SomethingPickOneName   <$> (HsJSONPB.parseField parseNameOrId_object "name")
    , Just . SomethingPickOneSomeid <$> (HsJSONPB.parseField parseNameOrId_object "someid")		
    , pure Nothing		
   ]
```

src/Proto3/Suite/DotProto/Generate.hs
themattchan: @ixmatus it comes from this function used when writing lisp macros: http://clhs.lisp.se/Body/f_gensym.htm

ixmatus: @themattchan oh okay.

src/Proto3/Suite/Class.hs
```diff
@@ -435,18 +437,20 @@ class MessageField a where
...
-                              Nothing -> [])
-                            Nothing
--- [todo] what were these intended for?
--- primDotProto :: DotProtoMessagePart -> DotProtoDefinition
--- primDotProto field = DotProtoMessage generateMessagePartName [ field ]
-
--- generateMessagePartName :: DotProtoIdentifier
--- generateMessagePartName = Single ""
+messageField ty packing = DotProtoField
+    { dotProtoFieldNumber = fieldNumber 1
```
Gabriel439: This looks so much nicer.  Thanks for little cleanups like this :)

PR #64, opened 2018-05-01
Add optional Dhall.Interpret and Dhall.Inject instances for generated protobuf Haskell types

ixmatus:
This change introduces an experimental and optional (off by-default) feature to generate [`Dhall.Interpret`](https://hackage.haskell.org/package/dhall-1.13.0/docs/Dhall.html#t:Interpret) and [`Dhall.Inject`](https://hackage.haskell.org/package/dhall-1.13.0/docs/Dhall.html#t:Inject) instances for Haskell types generated from parsed proto files.

The motivation for this change came, primarily, from the desire to have a "better JSONPB" because Dhall has static types provided with the data itself which makes it easier to read than JSON. Additionally, Dhall does not support "zero initialized values" which can make reading a Dhall rendering more informative and intuitive than a JSONPB rendering of the same message.

A additional change to automatically generate the Haskell source files from the test proto files was necessary in order to support building (and testing) this package with and without dhall.

###### Example

These protobuf types:

```protocol-buffer
message Trivial {
  int32 trivialField     = 1;
  AnotherTrivialType att = 2;
}

message AnotherTrivialType {
  string boopMe = 1;
}
```

... as Haskell data types:

```haskell
data Trivial = Trivial 
  { trivialTrivialField :: Int32
  , trivialAtt          :: Maybe AnotherTrivialType
  } deriving (Show, Eq, Ord, Generic)

data AnotherTrivialType = AnotherTrivialType
  { anotherTrivialTypeBoopMe :: Text 
  } deriving (Show, Eq, Ord, Generic)

let ex = Trivial 23 Nothing
```

... rendered to Dhall:

```dhall
{ trivialTrivialField =
    23
, trivialAtt =
    [] : Optional { anotherTrivialTypeBoopMe : Text }
}
```


proto3-suite.cabal
```diff
@@ -10,7 +10,17 @@ build-type:          Simple
 cabal-version:       >=1.10
 data-files:          test-files/*.bin tests/encode.sh tests/decode.sh
 
+Flag dhall
+  Description:   Turn on Dhall interpret and inject codegen
+  Default:       False
```
Gabriel439: You'll probably also want to add `Manual: True` so that `cabal` doesn't silently disable the `dhall` flag if it can't find a build plan

release.nix
```diff
@@ -11,12 +11,89 @@
...
+                          ghc = newHaskellPkgs.ghcWithPackages (pkgs:
+                            oldArgs.testHaskellDepends ++ [ proto3-suite-boot ]
+                          );
+                        in ''
+                          ${postPatch}
+
+                          export PATH=${newHaskellPkgs.cabal-install}/bin:${ghc}/bin:${python}/bin''${PATH:+:}$PATH
+                        '');
+
+                     # We need to add dhall as a dependency because
```
Gabriel439: I think `cabal2nix` will correctly add the `dhall` dependency if you do `cabal2nix --flag dhall`

proto3-suite.cabal
Gabriel439: See: https://www.haskell.org/cabal/users-guide/developing-packages.html#pkg-field-flag-manual

release.nix
ixmatus: Oh nice, I see it also generates a `configureFlags = ` definition...

PR #82, opened 2018-11-14
Make development a little bit easier and support GHC 8.4.x

ixmatus:
While working on building this repository with GHC 8.4.x I also decided to tackle the issue of needing to manually generate the Nix derivations for `proto3-suite` with cabal2nix so I switched to using `callCabal2nix`.

This necessitated cribbing a function that does not exist in the `nixpkgs-18.09` pin that I upgraded this repository to in order to support threading the `-fdhall` flag to `cabal2nix` when generating at run-time. This can go away once we advance the pin to 19.03 (or that function gets backported).

This change also updates the codebase for GHC 8.4.x.


j6carey:
Why 8.4.3 instead of 8.4.4?

ixmatus:
Because 8.4.4 is not in that specific revision of 18.09 (which I cribbed from our other software), I can bump the 18.09 pin to the latest though for 8.4.4

PR #91, opened 2019-03-30
Use Data.Coerce.coerce and unboxed proxy

themattchan:
This should speed up protobuf serde a bit (not sure by how much), by getting rid of the container traversals to wrap things in newtypes in `encodeMessage/decodeMessage`. 

Using unboxed proxy couples us to GHC, but that's ok because we only officially support GHC 8.0+. Because this changes the API for the `Named, Finite, Primitive` classes, the version has been bumped to `0.3.0.0`.

### DIFF
--- a/src/Proto3/Suite/Class.hs
+++ b/src/Proto3/Suite/Class.hs
@@ -60,6 +60,7 @@
 {-# LANGUAGE DefaultSignatures          #-}
 {-# LANGUAGE FlexibleContexts           #-}
 {-# LANGUAGE FlexibleInstances          #-}
+{-# LANGUAGE MagicHash                  #-}
 {-# LANGUAGE PolyKinds                  #-}
 {-# LANGUAGE ScopedTypeVariables        #-}
 {-# LANGUAGE TypeApplications           #-}
@@ -94,6 +95,7 @@ import           Control.Monad
 import qualified Data.ByteString        as B
 import qualified Data.ByteString.Base64 as B64
 import qualified Data.ByteString.Lazy   as BL
+import           Data.Coerce            (coerce)
 import           Data.Functor           (($>))
 import           Data.Int               (Int32, Int64)
 import qualified Data.Map               as M
@@ -106,7 +108,7 @@ import qualified Data.Text.Lazy         as TL
 import qualified Data.Traversable       as TR
 import           Data.Vector            (Vector)
 import           Data.Word              (Word32, Word64)
-import           GHC.Exts               (fromList)
+import           GHC.Exts               (fromList, Proxy#, proxy#)
 import           GHC.Generics
 import           GHC.TypeLits
 import           Proto3.Suite.DotProto  as DotProto
@@ -175,7 +177,7 @@ instance (Bounded e, Enum e) => HasDefault (Enumerated e) where
     case toEnumMay 0 of
       Nothing -> Enumerated (Left 0)
       Just x -> Enumerated (Right x)
-  isDefault = (== 0) . either id fromEnum . enumerated
+  isDefault = (== 0) . either id fromEnum . coerce @(Enumerated e) @(Either Int e)
 
 instance HasDefault (UnpackedVec a) where
   def = mempty
@@ -187,15 +189,15 @@ instance HasDefault (PackedVec a) where
 
 instance HasDefault (NestedVec a) where
   def = mempty
-  isDefault = null . nestedvec
+  isDefault = null . coerce @(NestedVec a) @(Vector a)
 
 instance HasDefault (Nested a) where
   def = Nested Nothing
-  isDefault = isNothing . nested
+  isDefault = isNothing . coerce @(Nested a) @(Maybe a)
 
 instance (HasDefault a) => HasDefault (ForceEmit a) where
   def       = ForceEmit def
-  isDefault = isDefault . forceEmit
+  isDefault = isDefault . coerce @(ForceEmit a) @a
 
 -- | Used in fields of generated records to represent an unwrapped
 -- 'PackedVec'/'UnpackedVec'
@@ -233,13 +235,13 @@ instance HasDefault (Fixed Int64)
 -- 'Generic' class, which will extract the name of the type constructor.
 class Named a where
   -- | Get the name of a type constructor
-  nameOf :: IsString string => Proxy a -> string
+  nameOf :: IsString string => Proxy# a -> string
 
-  default nameOf :: (IsString string, GenericNamed (Rep a)) => Proxy a -> string
-  nameOf _ = genericNameOf (Proxy @(Rep a))
+  default nameOf :: (IsString string, GenericNamed (Rep a)) => Proxy# a -> string
+  nameOf _ = genericNameOf (proxy# :: Proxy# (Rep a))
 
 class GenericNamed (f :: * -> *) where
-  genericNameOf :: IsString string => Proxy f -> string
+  genericNameOf :: IsString string => Proxy# f -> string
 
 instance Datatype d => GenericNamed (M1 D d f) where
   genericNameOf _ = fromString (datatypeName (undefined :: M1 D d f ()))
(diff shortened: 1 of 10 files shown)


themattchan:
Perhaps unboxed proxy should be controlled by a CPP flag?

src/Proto3/Suite/Class.hs
```diff
@@ -405,16 +407,16 @@ instance Primitive BL.ByteString where
   primType _ = Bytes
 
 instance forall e. (Bounded e, Named e, Enum e) => Primitive (Enumerated e) where
-  encodePrimitive num = Encode.enum num . enumify . enumerated
+  encodePrimitive num = Encode.enum num . enumify . coerce @(Enumerated e) @(Either Int e)
     where enumify (Left i) = i
           enumify (Right x) = fromEnum x
-  decodePrimitive = fmap Enumerated Decode.enum
-  primType _ = Named (Single (nameOf (Proxy @e)))
+  decodePrimitive = coerce @(_ (Either Int e)) @(_ (Enumerated e)) Decode.enum
```
ixmatus: I think partial type signatures in code make it more confusing to read; I can see why this is more concise when you have type-context in your head but without that context I have to do more work to figure out what those holes are. This follows the philosophy that types are documentation and explicit documentation will make the code easier to read (whether you're trying to maintain it or simply understand it).

For example, I have no idea what that type hole is and I'd be nervous to change this code without knowing it, so I stuck something I knew would give a type-error there just so I could see what the inferred type was (`Parser RawParser`). As an aside, I guess GHC doesn't provide type-hole information in type applications?

ixmatus: This comment applies to all of the partial type signature instances in the code.

ixmatus: About the point regarding conciseness: I understand that it feels fmappy by omitting the `Parser RawParser` but I think the partial signature produces more questions for the reader than the other way around. In fact I think the coerce is _nicer_ than the fmap version (setting aside convenience and possible efficiency reasons) from a type-based documentation perspective, if you fill in the partial type sigs.

src/Proto3/Suite/DotProto/Generate/Swagger.hs
```diff
@@ -70,9 +72,9 @@ ppSchema = LC8.putStrLn . encodePretty . toSchema
 -- | JSONPB schemas for protobuf enumerations
 instance (Finite e, Named e) => ToSchema (Enumerated e) where
   declareNamedSchema _ = do
-    let enumName        = nameOf (Proxy @e)
+    let enumName        = nameOf (proxy# :: Proxy# e)
```
j6carey: BTW, one trick is to replace `(proxy# :: Proxy# a)` witih `proxy# @a`.

src/Proto3/Suite/Class.hs
```diff
@@ -175,7 +177,7 @@ instance (Bounded e, Enum e) => HasDefault (Enumerated e) where
     case toEnumMay 0 of
       Nothing -> Enumerated (Left 0)
       Just x -> Enumerated (Right x)
-  isDefault = (== 0) . either id fromEnum . enumerated
+  isDefault = (== 0) . either id fromEnum . coerce @(Enumerated e) @(Either Int e)
```
j6carey: In this case I prefer the old syntax; the efficiency should be the same.

src/Proto3/Suite/Class.hs
```diff
@@ -405,16 +407,16 @@ instance Primitive BL.ByteString where
...
-  decodePrimitive = fmap Enumerated Decode.enum
-  primType _ = Named (Single (nameOf (Proxy @e)))
+  decodePrimitive = coerce @(_ (Either Int e)) @(_ (Enumerated e)) Decode.enum
+  primType _ = Named (Single (nameOf (proxy# :: Proxy# e)))
 
 instance (Primitive a) => Primitive (ForceEmit a) where
-  encodePrimitive num = encodePrimitive num . forceEmit
-  decodePrimitive     = fmap ForceEmit decodePrimitive
-  primType _          = primType (Proxy @a)
+  encodePrimitive num = encodePrimitive num . coerce @(ForceEmit a) @a
```
j6carey: As mentioned above, the original code is probably just as efficient as the new code.  I find the original code clearer.  If there is any benefit from using `coerce` in this case, which I doubt, then it would only arise if you coerced the entire function, not merely its argument.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -456,45 +458,145 @@ ctxtImports tyCtxt =
...
-hsTypeFromDotProtoPrim _    Float    = pure $ primType_ "Float"
-hsTypeFromDotProtoPrim _    Double   = pure $ primType_ "Double"
-hsTypeFromDotProtoPrim ctxt (Named msgName) =
+-- Convert a dot proto type to a Haskell type
+dptToHsType :: MonadError CompileError m => TypeContext -> DotProtoType -> m HsType
+dptToHsType = foldDPT dptToHsContType dpptToHsType
+
+-- Convert a dot proto type to a wrapped Haskell type
+dptToHsTypeWrapped :: MonadError CompileError m => [DotProtoOption] -> TypeContext -> DotProtoType -> m HsType
+dptToHsTypeWrapped opts =
```
j6carey: This looks a new feature not mentioned in your summary.  Could you augment your PR summary?

j6carey:
Should we increase the version number because the interface is changing?

src/Proto3/Suite/DotProto/Generate/Swagger.hs
themattchan: @j6carey type applications don't work with unboxed proxy for some reason:

```
src/Proto3/Suite/DotProto/Generate/Swagger.hs:75:35: error:
    • Couldn't match kind ‘e1’ with ‘*’
      ‘e1’ is a rigid type variable bound by
        the instance declaration
        at src/Proto3/Suite/DotProto/Generate/Swagger.hs:73:10-55
      When matching types
        a0 :: e
        a1 :: *
      Expected type: Proxy# a1
        Actual type: Proxy# a0
    • In the first argument of ‘nameOf’, namely ‘(proxy# @e)’
      In the expression: nameOf (proxy# @e)
      In an equation for ‘enumName’: enumName = nameOf (proxy# @e)
    • Relevant bindings include
        declareNamedSchema :: proxy (Enumerated e1)
                              -> Data.Swagger.Declare.Declare (Definitions Schema) NamedSchema
          (bound at src/Proto3/Suite/DotProto/Generate/Swagger.hs:74:3)
   |
75 |     let enumName        = nameOf (proxy# @e)
   |                                   ^^^^^^^^^
```

themattchan:
@j6carey the version has already been bumped, see summary

src/Proto3/Suite/DotProto/Generate.hs
themattchan: This isn't a feature but rather an implementation detail

src/Proto3/Suite/Class.hs
themattchan: I originally wrote the whole type out (see commit f815255), but I felt that it was too verbose and obscured the parts of the type that was actually different.

ixmatus: Yeah, as someone who may have to maintain this code, I prefer the verbose version you originally wrote.

themattchan: ok

src/Proto3/Suite/Class.hs
themattchan: Ok, I verified that the simplifier output is the same for both versions

themattchan: reverted in  9309ab1

src/Proto3/Suite/Class.hs
```diff
@@ -187,15 +189,15 @@ instance HasDefault (PackedVec a) where
 
 instance HasDefault (NestedVec a) where
   def = mempty
-  isDefault = null . nestedvec
+  isDefault = null . coerce @(NestedVec a) @(Vector a)
```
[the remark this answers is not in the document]
themattchan: reverted in 9309ab1

src/Proto3/Suite/Class.hs
themattchan: reverted in 9309ab1

themattchan:
@ixmatus that's right, if the compiler can `coerce` two types then it emits a `cast` core instruction to convert them (which does no actual work)

src/Proto3/Suite/DotProto/Generate/Swagger.hs
j6carey: Oh, right; the first type variable is probably the kind, so you would need to do `proxy# @_ @e` to go this route.

themattchan: As we discussed offline, I'm going to keep the existing code as type ascription is less confusing to read than the double type application version.

themattchan:
@j6carey and @ixmatus, thanks for the review.

ixmatus:
@themattchan no problem 😄

j6carey:
@themattchan , you're welcome.

src/Proto3/Suite/DotProto/Generate/Swagger.hs
j6carey: That's fine.

PR #95, opened 2019-04-15
Use the Applicative instance for the Parser

ndmitchell:
Means that we never go near the Monad bits, so could potentially switch parser, or use faster instances.


Gabriel439:
Thanks! 🙂

Issue #97, opened 2019-04-16
Invalid Haskell generated


ndmitchell:
Given:

```

syntax = "proto3";
package daml_lf_1;

message Foo {
  oneof Bar {
    sfixed64 Baz = 5;
  }
}
```

This generates:

```
((Hs.coerce @(Hs.Int64)
                                         @(HsProtobuf.Signed (HsProtobuf.Fixed Hs.Int64)))
                                        (HsProtobuf.ForceEmit y))
```

Which is ill-typed as the second argument is `ForceEmit Int64`, not `Int64`, leading to the error:

```
gen\DamlLf1.hs:52:42: error:
    * Couldn't match expected type `Hs.Int64'
                  with actual type `HsProtobuf.ForceEmit Hs.Int64'
    * In the third argument of `Hs.coerce', namely
        `(HsProtobuf.ForceEmit y)'
      In the second argument of `HsProtobuf.encodeMessageField', namely
        `((Hs.coerce
             @(Hs.Int64) @(HsProtobuf.Signed (HsProtobuf.Fixed Hs.Int64)))
            (HsProtobuf.ForceEmit y))'
      In the expression:
        (HsProtobuf.encodeMessageField
           (HsProtobuf.FieldNumber 5)
           ((Hs.coerce
               @(Hs.Int64) @(HsProtobuf.Signed (HsProtobuf.Fixed Hs.Int64)))
              (HsProtobuf.ForceEmit y)))
   |
52 |                [case fooBar of
   |                                          ^^^^^^^^^^^^^^^^^^^^^^
```

ndmitchell:
For a more complete example, which is what I was using as souce, see https://github.com/digital-asset/daml/blob/master/daml-lf/archive/da/daml_lf_1.proto and the 3 other files in that directory.

themattchan:
Resolved by #99

PR #99, opened 2019-04-17
#97, make sure OneofSubfield generates well-typed code for non-Maybe

ndmitchell:
Before the test case in #97 generated:

```
((Hs.coerce @(Hs.Int64)
   @(HsProtobuf.Signed (HsProtobuf.Fixed Hs.Int64)))
   (HsProtobuf.ForceEmit y))
```

Now it generates:

```
HsProtobuf.ForceEmit
    ((Hs.coerce @(Hs.Int64)
        @(HsProtobuf.Signed (HsProtobuf.Fixed Hs.Int64)))
    y)
```

I'm assuming the ForceEmit is important, and that seemed like the only sensible place to put it. But I was totally guessing... I welcome feedback!

### DIFF
--- a/src/Proto3/Suite/DotProto/Generate.hs
+++ b/src/Proto3/Suite/DotProto/Generate.hs
@@ -868,17 +868,18 @@ messageInstD ctxt parentIdent msgIdent messageParts = do
                       --    Constructor y -> encodeMessageField num (Nested (Just y)) -- for embedded messages
                       --    Constructor y -> encodeMessageField num (ForceEmit y)     -- for everything else
                       let mkAlt (OneofSubfield fieldNum conName _ dpType options) = do
-                            let wrapMaybe
+                            let isMaybe
                                    | Prim (Named tyName) <- dpType
-                                   , Just DotProtoKindMessage <- dotProtoTypeInfoKind <$> M.lookup tyName ctxt
-                                   = HsParen . HsApp (HsVar (haskellName "Just"))
+                                   = Just DotProtoKindMessage == fmap dotProtoTypeInfoKind (M.lookup tyName ctxt)
                                    | otherwise
-                                   = forceEmitE
+                                   = False
 
-                            xE <- wrapE ctxt options dpType
-                                   . wrapMaybe
+                            let wrapJust = HsParen . HsApp (HsVar (haskellName "Just"))
+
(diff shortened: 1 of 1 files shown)


src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -868,17 +868,18 @@ messageInstD ctxt parentIdent msgIdent messageParts = do
...
-                                   = forceEmitE
+                                   = False
 
-                            xE <- wrapE ctxt options dpType
-                                   . wrapMaybe
-                                   $ HsVar (unqual_ "y")
+                            let wrapJust = HsParen . HsApp (HsVar (haskellName "Just"))
 
+                            xE <- (if isMaybe then id else fmap forceEmitE)
+                                   . wrapE ctxt options dpType
```
themattchan: Using `dimap` would be nice here.

ndmitchell: Sorry, I don't see how to use dimap at all. Any hints?

ndmitchell:
Thanks for the review. Yep, I tried to do that coerce you mention first, but that requires adding ForceEmit to the type data type, which seemed to be not very clean.

src/Proto3/Suite/DotProto/Generate.hs
themattchan: Oops, you're right -- I thought both the pre- and post- operations were enabled by `isMaybe`.

This might read better though:

```
let adapt | isMaybe = lmap wrapJust
                | otherwise = rmap (fmap forceEmitE)

xE <- adapt (wrapE ctxt options dpType) (HsVar (unqual_ "y"))
```

ndmitchell: I find that harder to read. I guess I'd usually `(. wrapJust)` and `(fmap forceEmitE .)` rather than lmap/rmap - and in this case I'm not sure it's clearer at all. But happy to go that way if you want.

themattchan: I was thinking that it would be better to consolidate the adjustments predicated on `isMaybe`, but on second thought the extra indirection obfuscates matters.

themattchan:
Thanks for the bug fix, @ndmitchell!

PR #104, opened 2019-06-09
Refactor DotProto.Generate

themattchan:
This PR moves the non-Haskell specific code generator utilities from `DotProto.Generate` to `DotProto.Internal`. I've gone through the `DotProto.Generate` module and made stylistic improvements---the changes are largely cosmetic. This refactor also paves the way for a Purescript code generator feature that I'm hacking on.

Major changes: 

- Pass args struct to `compileDotProtoFile` runner
- Enforce that `Path` is nonempty using `List.NonEmpty`. This removes a case from `CompileError`
- Nice combinators for writing `importDecl_` statements (see `defaultImports`)
- Pass file name to parsec; report file name in parser errors
- Make `DotProtoServiceRPC` args a separate type
- Clarify codegen steps using `foldMapM` and `foldMapOfM`. This might also improve efficiency by causing separate traversal steps (e.g. map -> sequence -> mconcat) to be done in a single pass, depending on what ghc was doing with the original code...
- Improve readability and organisation of large codegen functions

The version has been updated to `0.3.0.1` to reflect minor changes in types and dependencies.


src/Proto3/Suite/DotProto/Generate/Common.hs
```diff
@@ -0,0 +1,358 @@
...
+
+qualifiedMessageName :: MonadError CompileError m => DotProtoIdentifier -> DotProtoIdentifier -> m String
+qualifiedMessageName parentIdent msgIdent = nestedTypeName parentIdent =<< dpIdentUnqualName msgIdent
+
+--------------------------------------------------------------------------------
+--
+-- ** Codegen bookkeeping helpers
+--
+
+-- | Bookeeping for qualified fields
```
themattchan: I think a lot of this can be removed, it's basically the same as `AST.DotProtoField`

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -151,24 +179,370 @@ data FindProtoResult
...
+dpIdentUnqualName :: MonadError CompileError m => DotProtoIdentifier -> m String
+dpIdentUnqualName (Single name)       = pure name
+dpIdentUnqualName (Dots (Path names)) = pure (last names)
+dpIdentUnqualName (Qualified _ next)  = dpIdentUnqualName next
+dpIdentUnqualName Anonymous           = internalError "dpIdentUnqualName: Anonymous"
+
+dpIdentQualName :: MonadError CompileError m => DotProtoIdentifier -> m String
+dpIdentQualName (Single name)       = pure name
+dpIdentQualName (Dots (Path names)) = pure (intercalate "." names)
+dpIdentQualName (Qualified _ _)     = internalError "dpIdentQualName: Qualified"
```
themattchan: This looks like a bug... shouldn't it be `Qualified q n = dpIdentQualName q <> "." <> dpIdentQualName n`? (ignoring the monad for illustrative purposes)

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -721,250 +515,256 @@ namedInstD messageName =
...
+          DotProtoMessageField fld     -> [dotProtoFieldName fld]
+          DotProtoMessageOneOf ident _ -> [ident]
+          _                            -> []
 
-           messagePartFieldD (DotProtoMessageOneOf fieldName _) = do
-               fullName <- prefixedFieldName messageName =<< dpIdentUnqualName fieldName
-               qualTyName <- prefixedConName messageName =<< dpIdentUnqualName fieldName
-               let fullTy = HsTyApp (HsTyCon (haskellName "Maybe")) . type_ $ qualTyName
-               pure [ ([HsIdent fullName], HsUnBangedTy fullTy) ]
+    foldMapM id
```
Gabriel439: 👍

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -1539,115 +1187,114 @@ dotProtoEnumD
...
+                               (HsUnGuardedRhs
+                                  (apply toEnumErrorE [enumNameE , uvar_ "i" , boundsE]))
+                               []
+      succFailure     = match_ (HsIdent "succ") [ HsPWildCard ]
+                               (HsUnGuardedRhs (HsApp succErrorE enumNameE)) []
+      predFailure     = match_ (HsIdent "pred") [ HsPWildCard ]
+                               (HsUnGuardedRhs (HsApp predErrorE enumNameE)) []
+
+      parseJSONPBDecls :: [HsMatch]
+      parseJSONPBDecls = foldr ((:) . matchConName) [mismatch] enumConNames
```
Gabriel439: Maybe `map matchConName enumConNames <> [ mismatch ]`

src/Proto3/Suite/DotProto/Internal.hs
Gabriel439: Probably.  I suggest adding a test case for the `.proto` code that this will fail on

tests/TestCodeGen.hs
```diff
@@ -113,10 +130,12 @@ simpleDecodeDotProto =
 
 -- E.g. dumpAST ["test-files"] "test_proto.proto"
 dumpAST :: [FilePath] -> FilePath -> IO ()
-dumpAST incs fp = do
-  Right (dp, tc) <- readDotProtoWithContext incs fp
-  let Right src = renderHsModuleForDotProto mempty dp tc
-  putStrLn src
+dumpAST incs fp = act >>= either (putStrLn . show) putStrLn
```
Gabriel439: Maybe use something like `throws` here, so that if the result is a `Left` the message is printed to `stderr` instead of `stdout`.  i.e.:

```haskell
dumpAST incs fp = throws act >>= putStrLn
```

src/Proto3/Suite/DotProto/Generate.hs
themattchan: the `foldr` is more efficient, though it doesn't matter here...

tests/TestCodeGen.hs
themattchan: I couldn't find any package that defines `throws`, so I just changed it to `either throwIO putStrLn`

src/Proto3/Suite/DotProto/Internal.hs
themattchan: So the only use of this function is this line in `dotProtoServiceD`

```
     packageName <- dpIdentQualName pkgIdent
```

which makes a string name from the package identifier given as a `DotProtoidentifier`... I guess that the package name can never be `Qualified`, so this code is fine?

We should look into straightening out this `DotProtoIdentifier` name resolution mess, but I'd prefer to do it in a later PR.

Issue #30, opened 2017-09-08
Ensure desired CG rendering of fixed scalar types


intractable:
In generated Haskell modules, we currently render record field types corresponding to protobuf message fields without any newtype wrapping, deferring necessary un/wrapping to the message's codec implementation. E.g., a field member for a nested message is rendered as the underlying `Maybe` type instead its contextually-correct `Nested` newtype wrapper.

The exception to this is the rendering for the fixed scalar types: `fixed32`, `sfixed32`, `fixed64`, and `sfixed64`. E.g., a `fixed32` field is currently rendered as `Fixed Word32`, in a data structure exposed to the consumer of the generated code – something we'd like to avoid.

The task here is to investigate whether or not this asymmetry is for some particular purpose, or if it is merely an historical implementation artifact that. If the former, let's document the snowflake, and if the latter (which is most likely), let's fix it.

Gabriel439:
I'm pretty sure that this is a legacy of generating `.proto` from Haskell and is not necessary for generating Haskell from `.proto`

intractable:
Ah, that makes sense. Thanks. Will leave this open so I remember to drop the `Fixed` wrapper.

themattchan:
Fixed in #92

PR #106, opened 2019-06-15
Generic HasDefault instances

themattchan:
This PR makes the code generator emit a `HasDefault` instance for every proto message and enum, which is computed using `GHC.Generics`. 

(It seems that the omission of this logical implementation was an oversight in the original implementation of `HasDefault`. Furthermore, this feature would allow us to remove a lot of `HasDefault` boilerplate from our internal projects.)

Example

```
* > data Foo = Foo { foo1 :: Int32, foo2 :: Maybe TL.Text } deriving (Show, Eq, Generic)
* > data Bar = Bar { boo1 :: Foo } deriving (Show, Eq, Generic)
* > instance HasDefault Foo
* > instance HasDefault Bar
* > def @Bar
Bar {boo1 = Foo {foo1 = 0, foo2 = Nothing}}
* > data X = A | B | C deriving (Show, Eq, Generic)
* > instance HasDefault X
* > def :: X
A
```

TODO

- [x] fix case for sum types
```
> data X = A | B | C deriving (Show, Eq, Generic)
* > instance HasDefault X

<interactive>:48:10: error:
    • No instance for (Proto3.Suite.Class.GenericHasDefault U1)
        arising from a use of ‘Proto3.Suite.Class.$dmdef’
    • In the expression: Proto3.Suite.Class.$dmdef @(X)
      In an equation for ‘def’: def = Proto3.Suite.Class.$dmdef @(X)
      In the instance declaration for ‘HasDefault X’
```

- [x] add some tests


src/Proto3/Suite/Class.hs
```diff
@@ -215,20 +224,20 @@ instance HasDefault (M.Map k v) where
   def = M.empty
   isDefault = M.null
 
--- TODO: Determine if we have a reason for rendering fixed32/sfixed as Fixed
```
themattchan: #30 was fixed by #92

PR #109, opened 2019-06-17
Render comments for messages, enums, and services.

j6carey:
Also, replace `Maybe String` with `String` as the way we note
comments in the AST.  The empty string renders as the absence of
a comment, whereas "\n" renders as an empty single-line comment.

Both of these changes modify the AST; therefore
we go to a new version of proto3-suite: 0.4.0.0.

This change also includes adaptations to support
multiple versions of GHC.

### DIFF
--- a/src/Proto3/Suite/DotProto/Generate.hs
+++ b/src/Proto3/Suite/DotProto/Generate.hs
@@ -694,7 +694,7 @@ messageInstD ctxt parentIdent msgIdent messageParts = do
                         , dpTypeE dotProtoFieldType
                         , dpIdentE dotProtoFieldName
                         , HsList (map optionE dotProtoFieldOptions)
-                        , maybeE (HsLit . HsString) dotProtoFieldComment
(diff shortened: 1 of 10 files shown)


src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -694,7 +694,7 @@ messageInstD ctxt parentIdent msgIdent messageParts = do
                         , dpTypeE dotProtoFieldType
                         , dpIdentE dotProtoFieldName
                         , HsList (map optionE dotProtoFieldOptions)
-                        , maybeE (HsLit . HsString) dotProtoFieldComment
+                        , HsLit (HsString dotProtoFieldComment)
```
themattchan: Use `str_`?

j6carey: Sounds good; thanks.  Changed in 15327e505a1f8cad11f2e98c8499a0a584a659ec .

j6carey:
@Gabriel439 , thanks for the review.

PR #111, opened 2019-07-11
Correctly handle gaps in enum codes.

j6carey:


src/Proto3/Suite/Tutorial.hs
```diff
@@ -92,21 +93,25 @@ instance Named Bar
...
+-- 'Finite', and 'Named' classes.  Each of these instances are implied by
+-- a 'Generic' instance, so can be derived as follows:
 --
 -- > data Shape
 -- >   = Circle
 -- >   | Square
 -- >   | Triangle
 -- >   deriving (Generic, Enum, Finite, Named)
+-- >
+-- > instance ProtoEnum Shape
```
Gabriel439: If you use `DeriveAnyClass` you can add `ProtoEnum` to the `deriving` clause.  In fact, it appears that the derived instances for `Finite` and `Named` are already assuming the use of `DeriveAnyClass` because they would not work otherwise

j6carey: @Gabriel439 , thanks; fixed by 2deefae0e25583ba203772a5c8ea6f08953549be .

.travis.yml
```diff
@@ -1,21 +1,19 @@
 language: nix
 
-script: nix-build --attr proto3-suite-linux --arg enableDhall true release.nix
+script: |
+  DERIVATION=$(nix-instantiate --attr proto3-suite-linux --arg enableDhall true release.nix)
```
themattchan: Thanks for fixing this!

j6carey:
@Acentelles , good point; thanks.  Updated in cd86818fdf8f9459ef141c1e3254c7797c1c8a53 .

j6carey:
Thanks to @themattchan , @Acentelles , and @Gabriel439 for their help with this PR.

Gabriel439:
@j6carey: You're welcome! 🙂

PR #120, opened 2020-02-04
GHC 8.8. support

cocreature:
This includes version bumps and a `MonadFail` instance for `ProtoParser`.


Gabriel439:
@cocreature: Thank you! 🙂

PR #124, opened 2020-03-31
Bump LTS version to 15.5

DeepakKapiswe:


Gabriel439:
I think you can fix the CI failure by changing `import Control.Monad` to use an explicit import list

DeepakKapiswe:
Fixed @Gabriel439 :)

Gabriel439:
@DeepakKapiswe: Thank you! 🙂

PR #131, opened 2020-06-19
Update Generate.hs to produce correct Haskell types for .proto wrapper values

rashadg1030:
To see the generated code:

- Build project with nix
- `cd` into `./protoToHsGenTest`
- Then run the command:
`result/bin/compile-proto-file --out gen --includeDir test-files --proto wrap.proto`

This will generate a `Wrap.hs` file.


src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -472,6 +473,19 @@ dpptToHsType ctxt = \case
...
+    "Int32Value"  -> pure $ primType_ "Int32"
+    "Int64Value"  -> pure $ primType_ "Int64"
+    "UInt32Value" -> pure $ primType_ "Word32"
+    "UInt64Value" -> pure $ primType_ "Word64"
+    "StringValue" -> pure $ primType_ "Text"
+    "BytesValue"  -> pure $ primType_ "Bytestring"
+    "BoolValue"   -> pure $ primType_ "Bool"
+    "FloatValue"  -> pure $ primType_ "Float"
+    "DoubleValue" -> pure $ primType_ "Double"
+    _             -> fail "Unknown Value"
```
Gabriel439, line 488: GHC has a nice feature where if no guard for a given pattern match succeeds, then it will fall through to the next pattern match.  That mean that you can restructure the match like this and you don't need to handle the case where it doesn't match any of the known values:

```suggestion
  Named (Dots (Path ("google" :| ["protobuf", x])))
    | x == "Int32Value"  -> pure $ primType_ "Int32"
    | x == "Int64Value"  -> pure $ primType_ "Int64"
    | x == "UInt32Value" -> pure $ primType_ "Word32"
    | x == "UInt64Value" -> pure $ primType_ "Word64"
    | x == "StringValue" -> pure $ primType_ "Text"
    | x == "BytesValue"  -> pure $ primType_ "Bytestring"
    | x == "BoolValue"   -> pure $ primType_ "Bool"
    | x == "FloatValue"  -> pure $ primType_ "Float"
    | x == "DoubleValue" -> pure $ primType_ "Double"
```

protoToHsGenTest/Unwrap.hs
```diff
@@ -0,0 +1,273 @@
+{-# LANGUAGE DeriveGeneric     #-}
```
Gabriel439, line 1: Did you mean to check in this generated file?

rashadg1030: Yes, this one. Want to makes sure the generated Haskell module is correct.

Gabriel439: I usually try to avoid including generated files in version control.  The reason why is similar in spirit to the Haskell idea of "make invalid states unrepresentable".  When one file in a repository is generated from another file within the same repository then there is the potential for the two files to get out of sync (for example: the original file is updated, but then people forget to regenerate the generated file from the original file).  Instead, I usually prefer to tests to create generated files "on the fly" as they are needed, to prevent this from happening

rashadg1030:
I'm trying to filter the list of `HsImportDecl` to get rid of the generated `Google.Protobuf.Wrappers` import, but for some reason that escapes me it still shows up in the generated Haskell file. I filter the list of imports with the predicate that the module name is not equal to `Google.Protobuf.Wrappers`. To see what's happening, I even tried filtering imports that were qualified, but it still had no effect on the generated file. I'm assuming I'm looking at the wrong place. I'm pretty sure something in `Generate.hs` needs to be changed, but I'm not sure why filtering the import list isn't working.

Gabriel439:
@rashadg1030: I just tested locally using your branch and it looks like the import filtering worked.  Maybe you are still looking at a stale version of the file?

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -472,6 +478,18 @@ dpptToHsType ctxt = \case
...
   Double   -> pure $ primType_ "Double"
+  -- For properly handling wrapper values found at:
+  -- https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto
+  Named (Dots (Path ("google" :| ["protobuf", x])))
+    | x == "Int32Value"  -> pure $ primType_ "Int32"
+    | x == "Int64Value"  -> pure $ primType_ "Int64"
+    | x == "UInt32Value" -> pure $ primType_ "Word32"
+    | x == "UInt64Value" -> pure $ primType_ "Word64"
+    | x == "StringValue" -> pure $ primType_ "Text"
+    | x == "BytesValue"  -> pure $ primType_ "Bytestring"
```
Gabriel439, line 489:
```suggestion
    | x == "BytesValue"  -> pure $ primType_ "ByteString"
```

rashadg1030:
> @rashadg1030: I just tested locally using your branch and it looks like the import filtering worked. Maybe you are still looking at a stale version of the file?

Haha got it. I thought I re-built the executable and everything. Let me try again. Thank you!

Gabriel439:
@rashadg1030: You're welcome! 🙂

Gabriel439:
@rashadg1030: I believe the generated code still does not type-check.  Could you ask @lambdadog to help you work through the remaining issues since she recently worked on something similar

lambdadog:
Putting the results of me and @rashadg1030's pair programming session here so that it's reference-able:

  1. We came to the decision that it's not reasonable (at least for now) to remove the Google.Protobuf.Wrappers
    - The reasoning for this is twofold:
      1. It will make code generation much more complex and non-obvious
      2. It would cause the generated code to not reflect the specification (the .proto file), which causes more complexities than it fixes -- The .proto file will require the wrappers import, so it's not that valuable or reasonable to make the generated Haskell not require it.
  2. The simplest and most obvious step forward is to just translate (as an example) `Hs.Maybe Hs.Int32` to `Hs.Maybe Google.Protobuf.Wrappers.Int32Value` before performing the coercion to `HsProtobuf.Nested Google.Protobuf.Wrappers.Int32Value` (see below) and keep the `encodeMessage` implementation otherwise the same.

A brief example of what this change (diffed from the non-unwrapped code) would look like in `encodeMessage`:  
```diff
 (HsProtobuf.encodeMessageField (HsProtobuf.FieldNumber 1)
   (Hs.coerce @(Hs.Maybe Google.Protobuf.Wrappers.Int32Value)
     @(HsProtobuf.Nested Google.Protobuf.Wrappers.Int32Value)
-    testAaa))
+    (testAaa >>= Google.Protobuf.Wrappers.Int32Value)))
```

## Addendum

`decodeMessage` would also need modification, with the Value-wrapped fields needing unwrapping, but `dotProto` should be able to stay the same.

tools/compile-proto-file/Main.hs
```diff
@@ -38,6 +38,9 @@ parseArgs = info (helper <*> parser) (fullDesc <> progDesc "Compiles a .proto fi
         <> metavar "DIR"
         <> help "Output directory path where generated Haskell modules will be written (directory is created if it does not exist; note that files in the output directory may be overwritten!)"
 
+    unwrap = switch $
+      long "unwrap"
+        <> help "If this flag is set, Protobuffer types are properly translated into Haskell Maybe types."
```
intractable, line 43: Minor suggestion: this help text may need to provide a little more context. A user unfamiliar with the history of the tooling I might think "why _wouldn't_ I want to have the proper translation?"

In fact, since this change will entail a package version bump anyway, I think it might make sense to have the new behavior be the default, with the flag provided to opt into the old behavior. If you decide to go that way, perhaps the help text phrasing for the `--use-legacy-wrapper-types` option (e.g.) could be something like "Use legacy newtype wrapping for protobuf wrapper types (e.g. Int32Value) in generated Haskell code (provided for backwards compatibility)" or something along those lines.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -135,14 +137,17 @@ renderHsModuleForDotProto extraInstanceFiles dotProto importCtxt = do
 -- Instances given in @eis@ override those otherwise generated.
 hsModuleForDotProto
     :: MonadError CompileError m
-    => ([HsImportDecl], [HsDecl])
+    => Bool
+    -- ^
```
intractable, line 141: Missing a comment here. Might also be worth a `UnwrapWrapperTypes` newtype or somesuch in order to avoid boolean blindness at call sites.

tools/compile-proto-file/Main.hs
rashadg1030: Okay got it. I'm just reading this comment, but I will implement this in my next commit. I like this idea.

src/Proto3/Suite/DotProto/Generate.hs
rashadg1030: Yes, I agree that wrapping the `Bool` in a newtype is a good idea. Will do this.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -66,14 +68,73 @@ import           Turtle                         (FilePath)
...
+
+data Test = Test
+  { testX :: Maybe Int32
+  , testY :: Maybe ByteString
+  } deriving (...)
+
+...
+@
+-}
+newtype UseLegacyTypes = UseLegacyTypes { getBool :: Bool }
```
Gabriel439, line 136: Minor suggestion: Instead of using `Bool`, perhaps name the constructors after what the user is selecting, like this:

```haskell
data UseWrappers = Wrappers | NoWrappers
```

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -472,13 +537,50 @@ dpptToHsType ctxt = \case
   Bool     -> pure $ primType_ "Bool"
   Float    -> pure $ primType_ "Float"
   Double   -> pure $ primType_ "Double"
+  {-| For properly handling wrapper values found at:
+      <https://github.com/protocolbuffers/protobuf/blob/master/src/google/protobuf/wrappers.proto>
+      This functionality is new for now and hides behind the --unwrap flag.
+  -}
+  Named (Dots (Path ("google" :| ["protobuf", x])))
+    | x == "Int32Value"  -> pure $ primType_ "Int32"
```
Gabriel439, line 545: Instead of creating two separate `dppToHsTypeOld` and `dppToHsType` functions, you can just add a check on each of these branches to make sure that legacy mode is not enabled.  That way you don't have to worry about code drift between the legacy and non-legacy code

rashadg1030: @Gabriel439 Got it! I will do this.

src/Proto3/Suite/DotProto/Generate.hs
rashadg1030: @Gabriel439 This looks better!

Gabriel439:
@rashadg1030: For the `dhall` test failures, we can fix those by disabling the test suite for now (since they are trying to connect to a remote host).  You can do that by adding `"dhall"` to this list:

https://github.com/awakesecurity/proto3-suite/blob/fd767cba2aeda9510b43404c352776249626ebf0/release.nix#L138-L144

rashadg1030:
@Gabriel439 Will do!

Gabriel439:
It looks like the test suite needs to be updated to match the change to the type of `renderHsModuleForDotProto`

rashadg1030:
@Gabriel439 got it doing that now. Also cleaning up old comments I was using for myself.

rashadg1030:
Fixed tests. I ran them on my side and they all passed

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -344,13 +408,13 @@ dhallPBName name = Qual (Module hsDhallPB) (HsIdent name)
 
 dhallInterpretInstDecl :: String -> HsDecl
 dhallInterpretInstDecl typeName =
-  instDecl_ (dhallPBName "FromDhall")
+  instDecl_ (dhallPBName "Interpret")
```
Gabriel439, line 411: The tests are failing because of this change.  This may be due to incorrectly resolving a merge conflict when merging `master` into your branch.  If you change this back to `FromDhall`/`ToDhall` then the test should no longer fail on that

rashadg1030: @Gabriel439 Yes I noticed this and though something is weird. I run `git add --patch` to review my changes before I stage them and this one made me wonder what I did. Thank you!

rashadg1030:
Still failing. Fixing.

rashadg1030:
Not sure how to correct the two scripts `encode.sh` and `decode.sh`. I get a file not found error so I tried adding `Wrap.hs` to those files, but still arises. Testing on my machine though, those two tests pass.

rashadg1030:
@Gabriel439 I think this branch is ready to be merged. I disabled the two tests that were making CI fail.

Gabriel439:
@rashadg1030: Awesome!  Thank you for doing this 🙂

PR #132, opened 2020-07-06
Add dhall to dontCheckOverrides

rashadg1030:


Gabriel439:
@rashadg1030: So it looks like there was a problem with CI after all 🙂 

I'll go ahead and merge this despite the failure since it's a step along the way towards fixing CI

rashadg1030:
@Gabriel439 Got it. Where should I go from here?

Gabriel439:
@rashadg1030: The first step is to to disable the two failing tests until we can fix them.  After that, create a branch that restores the tests so that they print additional debugging information if they fail (like the current working directory and the contents of the `./tests` subdirectory)

PR #134, opened 2020-07-11
Remove useLegacyTypes flag that guards old behavior of compile-to-proto

rashadg1030:
When adding the new behavior for translating `protobuf` files to Haskell modules, we added a flag to keep the old behavior for backwards compatibility (the new behavior became the default). In this PR we are now completely removing that flag and completely removing the old behavior.

### DIFF
--- a/src/Proto3/Suite/DotProto/Generate.hs
+++ b/src/Proto3/Suite/DotProto/Generate.hs
@@ -24,7 +24,6 @@ module Proto3.Suite.DotProto.Generate
   ( CompileError(..)
   , TypeContext
   , CompileArgs(..)
-  , UseLegacyTypes(..)
   , compileDotProtoFile
   , compileDotProtoFileOrDie
   , hsModuleForDotProto
@@ -89,7 +88,7 @@ compileDotProtoFile CompileArgs{..} = runExceptT $ do
 
     extraInstances <- foldMapM getExtraInstances extraInstanceFiles
 
-    haskellModule <- renderHsModuleForDotProto useLegacyTypes extraInstances dotProto importTypeContext
+    haskellModule <- renderHsModuleForDotProto extraInstances dotProto importTypeContext
 
     liftIO (writeFile (FP.encodeString modulePath) haskellModule)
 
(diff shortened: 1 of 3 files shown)


rashadg1030:
@Gabriel439 said that there will have to be changes made to the UI to accommodate the new behavior added in my last PR. I assume it's best if we remove the flag after that change on the frontend is made.

j6carey:
Do we want to release a version with the compatibility flag still in place, or just figure this is all a bug fix anyway?

rashadg1030:
> Do we want to release a version with the compatibility flag still in place, or just figure this is all a bug fix anyway?

@jcarey yes so currently the version of `proto3-suite` that we recently integrated into `end-to-end` has a flag for backwards compatibility, but @Gabriel439 told me we will be removing it completely. Maybe it's too soon though. We probably should make the changes to the UI first and then if everything goes right remove the flag.

Gabriel439:
@rashadg1030: I wouldn't remove the flag until we cut a release.  Ideally there should be a migration window, meaning a release published where both the old way and new way are supported so that people have a chance to adjust their code to the new way of doing things.  Then we can remove the old behavior and publish yet another release after that

rashadg1030:
@Gabriel439 Okay great. Is there anything else that I should do related to this PR? OR should I move on to something else. Is the next step fixing `grpc-rest` as we discussed before?

Gabriel439:
@rashadg1030: For this pull request specifically the next step is to first cut a release.  You can request a Hackage account here:

https://hackage.haskell.org/accounts

... and once you have such an account I can grant you permission to upload a new `proto3-suite` release to Hackage and show you how to do so

In parallel, you should create a pull request for our internal code to switch to the new wrapper-free behavior

rashadg1030:
@Gabriel439 Got it. Thank you!

Gabriel439:
@rashadg1030: You're welcome! 🙂

rashadg1030:
@Gabriel439 @j6carey Do I have the green light to squash and merge this PR?

PR #148, opened 2021-04-19
Simplify Nix build

Gabriel439:
This is essentially the same change as https://github.com/awakesecurity/proto3-wire/pull/67,
except for the `proto3-suite` project


release.nix
```diff
@@ -244,49 +122,25 @@ let
...
-
-    preferLocalBuild = true;
-  };
+  pkgs = import nixpkgs { inherit config overlays; };
 
   linuxPkgs =
     import nixpkgs { inherit config overlays; system = "x86_64-linux" ; };
 
   darwinPkgs =
     import nixpkgs { inherit config overlays; system = "x86_64-darwin"; };
```
evanrelf, line 142: Is it really necessary for us to create system-specific attributes, when the user could just specify `--system` at the command-line themselves?

evanrelf (approved):
YAY

PR #149, opened 2021-04-23
Disable missing exports list warning in generated file

TristanCacqueray:
This change prevents this ghc compilation warning: `The export item ‘module ...’ is missing an export list`.


Gabriel439:
Thank you! 🙂

PR #151, opened 2021-05-09
Repeated sints

functorious:
The haskell code generated for repeated sint{32,64} fields in a .proto
relies on `PackedVec (Signed Int{32,64})` being an instance of
`MessageField`.

Without this instance GHC cannot compile the code generated by `compile-proto-file` when it is used with a `.proto` file containing messages with `repeated sint{32,64}` fields.

There is some logic duplicate with proto3-wire (`zigZagEncode`, and `zigZagDecode` - this last just straight copied), but I wasn't sure what the process was for updating dependencies.


functorious:
This is actually *not* ready - I completely missed the other test groups when I was updating the tests, and apparently `decode . encode` is not always and identity operation!

functorious:
> This is actually _not_ ready - I completely missed the other test groups when I was updating the tests, and apparently `decode . encode` is not always an identity operation!

Okay, bug fixed. The comment for the commit and fix is interesting, the bug may exist in Proto3.Wire.Decode as well. I'm also not an expert in this area at all, so I could just not be understanding the issue. But the property test seems rather straightforward and a good guide on what is minimally correct.

Gabriel439 (approved):
If you would like to obtain the zigzag encoding/decoding logic from the `proto3-wire` package, you can update the dependency by installing `cabal2nix` and running:

```bash
$ # To update to the latest Hackage version:
$ cabal2nix cabal://proto3-wire > ./nix/proto3-wire.nix

$ # To update to `master`
$ cabal2nix https://github.com/awakesecurity/proto3-wire.git > ./nix/proto3-wire.nix
```

If not, we can merge this PR as is

evanrelf:
> ```
> $ # To update to the latest Hackage version:
> $ cabal2nix cabal://proto3-wire > ./nix/proto3-wire.nix
> ```

You should run `cabal update` before this to ensure your local package list matches the latest from Hackage.

functorious:
I've made a PR to `proto3-wire` to export the zigzag codec functions, if you're happy exporting them I'll remove the duplication here.

Gabriel439:
@functorious: Thank you for doing this!  I can upload this to Hackage if you don't have any other pending changes

functorious:
@Gabriel439 Glad to help! I don't have any more changes - cut a release whenever you'd like.

PR #156, opened 2021-06-05
Update stack.yaml to be compatible with latest stack versions

GulinSS:
See https://www.fpcomplete.com/blog/2017/07/stacks-new-extensible-snapshots/


Gabriel439:
Thank you! 🙂

PR #157, opened 2021-06-09
Switch from Travis CI to GitHub Actions

evanrelf:
Travis CI's Nix support seems a lot less mature compared to GitHub Actions. I prefer we standardize on using GitHub Actions with Cachix for our open source projects.

#### Todo

- [ ] Change required status in repo settings


Gabriel439 (approved):

lambdadog (commented):
Would it be useful to add a darwin builder as well, perhaps?

If we set up build to run on a matrix with `os: [macos-latest, ubuntu-latest]` and have the `runs-on` field use said value we could get darwin builds cached as well. I'm not 100% certain how useful it would be but it might be nice for developers at the very least.

lambdadog:
Of relevance would be the GitHub Actions usage limits, since they're tighter for the macos builders, but I don't thing we run a huge risk of hitting those and if we did it would just stall CI a little: https://docs.github.com/en/actions/reference/usage-limits-billing-and-administration#usage-limits

If we really wanted to avoid them we could look into making the darwin build conditional on the linux one (although we couldn't use a simple matrix then), but that would slow things down and is probably unnecessary.

evanrelf:
@lambdadog

> Would it be useful to add a darwin builder as well, perhaps?

I think a Darwin builder would be useful! We're missing documentation for how to take advantage of the caching, so we should also make sure to mention `awakesecurity.cachix.org` in the README or something.

> If we set up build to run on a matrix with os: [macos-latest, ubuntu-latest] and have the runs-on field use said value...

I added Darwin builds to `spectacle` [here](https://github.com/awakesecurity/spectacle/blob/main/.github/workflows/ci.yml#L23-L37) by just copy pasting the Linux job and replacing `ubuntu-latest` with `macos-10.15`. I wasn't aware of the matrix stuff; that seems a lot less boilerplate-y.

If you know how to get that working, could you push that to this branch?

> Of relevance would be the GitHub Actions usage limits, since they're tighter for the macos builders, but I don't think we run a huge risk of hitting those and if we did it would just stall CI a little

My understanding is GitHub just limits concurrent builds, and not total build time, at least on open source projects. We don't have very many PRs open at once, so I don't expect to hit any limits.

> we could look into making the darwin build conditional on the linux one

I also considered that for `spectacle`, however I think there's still benefit in caching transitive dependencies for Darwin even the build ultimately fails. For example, if you change a Nixpkgs pin and all the deps build fine, but the project doesn't build, it's nice to cache the deps so your Nix shell loads quickly.

ixmatus (approved):

PR #161, opened 2021-06-28
More Nix refactoring

evanrelf:
Commits are atomic.

### DIFF
--- a/shell.nix
+++ b/shell.nix
@@ -10,7 +10,7 @@
 , compiler ? "ghc884"
 , enableDhall ? false
 , enableSwagger ? true
-}@args:
+}:
 
 with (import ./default.nix { inherit compiler enableDhall enableSwagger; });
 
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -13,7 +13,6 @@ jobs:
           - macos-10.15
         compiler:
           - ghc884
-          - ghc8104
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2.3.4
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -11,6 +11,9 @@ jobs:
         os:
           - ubuntu-latest
           - macos-10.15
+        compiler:
+          - ghc884
+          - ghc8104
     runs-on: ${{ matrix.os }}
     steps:
       - uses: actions/checkout@v2.3.4
@@ -22,4 +25,8 @@ jobs:
         with:
           name: awakesecurity
           authToken: '${{ secrets.CACHIX_AUTH_TOKEN }}'
-      - run: nix-build --arg enableDhall true
+      - run: >-
+          nix-build
+            --attr proto3-suite
+            --argstr compiler "${{ matrix.compiler }}"
+            --arg enableDhall true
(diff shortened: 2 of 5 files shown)


shell.nix
```diff
@@ -1,20 +1,29 @@
...
 if fast then
-  proto3-suite-boot.env
+  builtins.trace
+    ''
+    Using `fast` shell... here are some gotchas to be aware of:
+      - You must run codegen for test protos manually using the
+        `compile-test-protos` shell script (in the shell's `PATH`)
+      - You can _build_ the test suite, but you can't _run_ it (without it failing
+        due to missing dependencies)
+    ''
```
lambdadog, line 26: I admit I'm a bit confused about this bit.

You have two bullet points, the first saying that you *must* manually run a command for the test suite to work, then the second says that the test suite *doesn't* work.

Could possibly use a rewording?

evanrelf: Yeah you're right, that isn't clear. You need to run that script to create Haskell modules required for building the test suite. But that's as far as you can go; you can't actually run it without it failing.

lambdadog: I guess I'm not entirely sure why you'd want to build the test suite if you're not going to run it.

evanrelf: Being able to build the test suite isn't a huge benefit if you can't run it, but I figured it was still an improvement.

evanrelf: Reworded things, hopefully it's clearer now. Ultimately I want to get rid of the bootstrapping process, and then the "fast" shell won't be necessary. But for now, I'm providing a "fast" shell and making the caveats clear.

evanrelf:
I've built this PR successfully so many times on `hydra-dev` and my Mac laptop... 
- I've made sure to use the same arguments as CI is for `default.nix`. I've made sure my Git worktree is clean.
- I've re-cloned from scratch.
- I've rebased against the latest changes on `master`.
- I've built the result of merging this branch into `master`.

Everything works locally, _nothing_ works in CI. CI keeps showing the same error, as though my changes haven't taken effect. I don't know what to do ☹️

Gabriel439:
@evanrelf: I'd suggest trying to copy those derivations (and their closures) to a local machine.  For example, see if you can push them to cachix or something similar.  Then compare the derivations against the ones you get locally using `nix-diff` and see why they don't match.

evanrelf:
@Gabriel439 I was able to SSH into the runner but `nix copy`, `scp`, `rsync`, and `cachix` all didn't seem to work, probably due to how the minimal Ubuntu image is configured.

For now I'm just going to remove the commits that cause problems. I can revisit them in a future PR if I still feel like they're worth including.

evanrelf:
I've stripped out the problematic bits (to be revisited in the future, maybe), so what's left now passes CI and is ready for review.

evanrelf:
@Gabriel439 @lambdadog boop

lambdadog (approved):
LGTM!

My only minor concern is that after removing 8.10 we'll never add it back. It might be better to use the default version in the current iteration of nixpkgs we're using so it updates automatically when we move to a newer version down the line?

evanrelf:
@lambdadog Before this PR we were only building with GHC 8.8 (non-default compiler for the Nixpkgs pin we have), so this PR isn't a regression or anything.

I am eager to move this repo to GHC 8.10 to stay up-to-date with GHC releases, and match what we use in `end-to-end`. I just haven't dug into it yet (gotta make sure `grpc` and `protobuf` stay on the same version as they are now).

PR #166, opened 2021-07-20
Upgrade nixpkgs to 21.05

riz0id:
As mentioned [here](https://github.com/awakesecurity/proto3-wire/pull/71), the pinned nixpkgs for `proto3-wire` made it impossible to `nix-build` or `nix-shell` on Big Sur 11.X. I updated the `proto3-wire` dependency here so that `proto3-suite` can be built on Big Sur 11 as well as tried my best to clean up some of the nix.


nix/haskell-packages.nix
```diff
@@ -0,0 +1,126 @@
...
+      pkgsOld.lib.composeExtensions
+        (old.overrides or (_: _: {}))
+        (haskellPackagesNew: haskellPackagesOld:
+          let
+            proto3-suite-flags = (if enableDhall then ["-fdhall"] else []) ++ (if enableSwagger then [] else ["-f-swagger"]);
+
+          in {
+            tasty = haskellPackagesNew.callPackage ./haskell-packages/tasty.nix { };
+
+            tasty-ant-xml = haskellPackagesNew.callPackage ./haskell-packages/tasty-ant-xml.nix { };
```
evanrelf, line 28: I would really like to avoid checking in `cabal2nix` codegen. Please use [`packageSourceOverrides`](https://github.com/NixOS/nixpkgs/blob/7e9b0dff974c89e070da1ad85713ff3c20b0ca97/pkgs/development/haskell-modules/lib.nix#L43-L53) in a new extension, or call [`callCabal2nix` or `callHackage`](https://github.com/NixOS/nixpkgs/blob/7e9b0dff974c89e070da1ad85713ff3c20b0ca97/pkgs/development/haskell-modules/make-package-set.nix#L182-L220) directly, instead of `callPackage` on generated Nix code.

nix/haskell-packages.nix
```diff
@@ -0,0 +1,126 @@
+{ compiler, enableDhall, enableSwagger }:
```
evanrelf, line 1: I'd prefer we standardize on putting overlays under `nix/overlays/`, so ideally this would be `nix/overlays/haskell-packages.nix`. But if you have a preference otherwise, let me know.

release.nix
```diff
@@ -0,0 +1,13 @@
+{ compiler ? "ghc884"
```
evanrelf, line 1: I'd prefer we move away from `release.nix` in our repos, and standardize on just using `default.nix`.

`default.nix` is the default, idiomatic Nix entrypoint. There isn't any value in symlinking `default.nix` to `release.nix` since we aren't using Hydra.

nix/haskell-packages.nix
```diff
@@ -0,0 +1,126 @@
...
+                  sha256 = "14r2qm6x4bcaywbi3cypriz4hr8i2v3j4qm61lal6x21p0z9i9ak";
+                };
+              in haskellPackagesNew.callCabal2nix "proto3-wire" source { };
+
+            proto3-suite-base =
+              haskellPackagesNew.callCabal2nixWithOptions
+                "proto3-suite"
+                (gitignoreSource ../.)
+                (pkgsNew.lib.concatStringsSep " " proto3-suite-flags)
+                { };
```
evanrelf, line 45: Why remove the `pname` and `configureFlags` overrides for `proto3-suite-base`?

I definitely want to keep the `pname` override. The `configureFlags` might not be necessary, but I'm not sure; was that something you tested?

If they aren't necessary, then I think you should inline `proto3-suite-flags` since it's only being used in one place.

evanrelf (commented):
The Nixpkgs pin bump and removal of redundant imports are great; thank you!

However, the other Nix changes you made are somewhat disruptive, and it's tricky to tell what all you changed since it's all in a single commit.

nix/haskell-packages.nix
riz0id: The removal of `pname` was not intentional but I did remove `configureFlags` in the attribute overrides since it is redundant.

nix/haskell-packages.nix
evanrelf: I would add this to the `haskellPackages` extension:

```nix
"range-set-list" = pkgsOld.haskell.lib.overrideCabal haskellPackagesOld."range-set-list" (old: {
  broken = false;
  jailbreak = true;
});
```

evanrelf (approved):

evanrelf:
You might want to change the default compiler from GHC 8.8.4 to GHC 8.10.4 (`grep -r ghc884`). We've been meaning to move to 8.10, and since it's the default compiler for Nixpkgs 21.05, it will speed up CI because it won't have to build GHC from scratch.

PR #167, opened 2021-07-20
Use Swagger schema `format` to distinguish wrapper types from primitives

evanrelf:
Notes:

- These `format` values are non-standard / incorrect Swagger, so they're hidden behind a flag (off by default).
- Replacing the original `instance ToSchema (OverrideToSchema ByteString)` is okay because [we don't have `optional` anymore](https://github.com/awakesecurity/proto3-suite/pull/165), therefore `Maybe ByteString` now unambiguously refers to `BytesValue`.

cc @natefaubion

Commits are atomic.

### DIFF
--- a/proto3-suite.cabal
+++ b/proto3-suite.cabal
@@ -32,6 +32,11 @@ flag swagger
   Default:       True
   Manual:        True
 
+flag swagger-wrapper-format
+  Description:   Change Swagger schema format for Protobuf wrapper types
+  Default:       False
+  Manual:        True
+
 library
 
   if flag(dhall)
@@ -43,6 +48,8 @@ library
     exposed-modules:   Proto3.Suite.DotProto.Generate.Swagger
     build-depends:     swagger2 >=2.1.6 && <2.7
     cpp-options:       -DSWAGGER
+    if flag(swagger-wrapper-format)
+      cpp-options:       -DSWAGGER_WRAPPER_FORMAT
 
   exposed-modules:     Proto3.Suite
                        Proto3.Suite.Class
@@ -112,6 +119,8 @@ test-suite tests
   if flag(swagger)
     build-depends:     swagger2
     cpp-options:       -DSWAGGER
+    if flag(swagger-wrapper-format)
+      cpp-options:       -DSWAGGER_WRAPPER_FORMAT
 
   other-modules:       ArbitraryGeneratedTestTypes
                        TestCodeGen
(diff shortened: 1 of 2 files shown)


proto3-suite.cabal
```diff
@@ -43,6 +48,8 @@ library
     exposed-modules:   Proto3.Suite.DotProto.Generate.Swagger
     build-depends:     swagger2 >=2.1.6 && <2.7
     cpp-options:       -DSWAGGER
+    if flag(swagger-wrapper-format)
+      cpp-options:       -DSWAGGER_WRAPPER_FORMAT
```
Gabriel439, line 52: Instead of using `CPP`, a cleaner approach is to move all of the code into a separate module, which is conditionally included in the `.cabal` file.  In other words:

```
if flag(swagger-wrapper-format)
  hs-source-dirs: swagger
```

The other benefit of doing it this way is that it also lets you conditionally depend on any necessary modules (not that we need that in this particular case)

evanrelf: Yeah you're right, in this case it would be a lot cleaner that way.

evanrelf: Hmm actually, I'm not so sure. I think I still need `-XCPP` for the conditional `import`. I'm inclined to stick with just `-XCPP`, and not `-XCPP` + a new module. As much as I hate `-XCPP`, in this case I think it's already as simple as it's going to get.

evanrelf: I think this is something better addressed in a future, more broad refactoring.

Gabriel439: You don't need a conditional import if you do it like this:

* Decide on the name for the separate module (e.g. `Proto3.Suite.DotProto.Generate.Instances`)
* Create two source directories: `instances` and `no-instances`
* Under `no-instances`, place a blank module of the desired name
* Under `instances`, place the module with the instances
* Use an `if` expression in the `.cabal` file to determine which source directory to import based on the configure flag
* `Proto3.Suite.DotProto.Generate.Swagger` unconditionally imports the desired module

evanrelf: Ah good point. I like that idea. Would you mind if I tackled that in a follow-up PR, so I can apply that technique to _all_ our flags?

Gabriel439: Alright, that's fine

Gabriel439 (approved):

lambdadog (approved):

PR #168, opened 2021-07-21
Reduce use of `-XCPP` for conditional compilation

evanrelf:
Commits are atomic


Gabriel439 (approved):
Thanks for doing this!

PR #180, opened 2022-02-13
Move AST to separate package

riz0id:
This PR isolates `Proto3.Suite.DotProto.AST` into separate package `proto3-suite-ast`.


evanrelf (approved):

Gabriel439:
The original request for this was from quite a while so I wanted to check if we still need to split this out into a separate package?

I do think there is a good reason to split into multiple packages, but for a different reason: I think there should be a package that implements the code generation logic and the generated code should depend on a separate package.  However, I don't believe separating out the AST into a separate package addresses that.

riz0id:
I can't think of a reason why the syntax datatype would be useful in a separate package even if I went ahead and added the parser into the new package. I don't mind removing the new package, but I do think that some of the changes in this PR like documentation, removing orphan instances out of `Proto3.Suite.DotProto.Rendering`, and the new modules compartmentalizing things a bit more are useful though.

PR #185, opened 2022-03-18
specify autogen-modules

fumieval:
Otherwise cabal build fails when used in source-repository-package

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,4 @@ test-files/test_files/
 # Ignore Emacs backup files
 *~
 # Ignore code generation artifacts
-gen/
+gen/*.hs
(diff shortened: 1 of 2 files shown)


Gabriel439:
See: https://github.com/awakesecurity/proto3-suite/pull/147

fumieval:
If the autogen-module section is missing, it simply fails to build with the following error (cabal-install version 3.6.2.0):

```
cabal: sdist of proto3-suite-0.5.0: Error: Could not find module: TestProto
with any suffix:
["gc","chs","hsc","x","y","ly","cpphs","hs","lhs","hsig","lhsig"]. If the
module is autogenerated it should be added to 'autogen-modules'.
```

Or is it unsupported to use this Git repository as an extra cabal.project dependency?

Gabriel439:
@fumieval: You might want to leave a comment on https://github.com/haskell/cabal/issues/7348, because that's where I was told not to use `autogen-modules` for this purpose

fumieval:
@Gabriella439 I found an IMHO reasonable workaround, which is adding a file to `gen` and include it in extra-source-files. It makes the package valid both in terms of Hackage and source-repository-package.

Gabriella439:
Thank you for following up on this! 🙂

PR #194, opened 2022-04-12
Break proto3-suite into 4 packages without cabal flags.

j6carey:
Rather than using cabal flags to control optional support for
Dhall and Swagger, along with associated library dependencies,
this change breaks proto3-suite into the following four packages:

  proto3-compile: compile-proto-file and protobuf AST

  proto3-base: supporting library for all generated code

  proto3-dhall: supporting library for generated Dhall integration

  proto3-swagger: supporting library for generated Swagger integration

The compile-proto-file executable now has command-line options
"--enableDhall" and "--enableSwagger" to trigger generation of
code to integrate with Dhall and/or Swagger.
  
This change consolidates the confusingly-named modules that
supported code generated for Swagger into a new module named
Proto3.Suite.JSONPB.Swagger.
    
For simplicity, this commit drops any attempt to support an integration
with Swagger that omits schema overrides for wrapper types.

Questions for review include:
1. What to do about `CHANGELOG.md`?
2. What to do about `README.md`?
3. How to test `Dockerfile`?


gbaz:
should one ( likely -compile ? ) still be named proto3-suite for continuity? (especially concerning the package name on hackage)

j6carey:
@gbaz , I had mixed feelings about it, because it would cause confusion between the package and the source control repository.  Also, none of the parts is really a "suite" of things.

But I have no strong opinion and will go along with the consensus.  @Gabriel439 , any thoughts on that?

evanrelf:
> should one ( likely -compile ? ) still be named proto3-suite for continuity? (especially concerning the package name on hackage)

I think the "suite" name only makes sense if you're including the full suite of functionality. None of the packages John broke out really fit the "suite" name on their own.

Considering this is already a large breaking change, I think either deprecating the `proto3-suite` package, or repurposing it as purely re-exports of the new focused packages, is warranted.

Gabriel439:
I think it would help if there were one release of `proto3-suite` that re-exported the same API (or as close to the old API as possible), that way it's easier for end users to migrate.  Then we could stop publishing new releases for `proto3-suite` and instead focus on the underlying packages.

j6carey:
@Gabriel439 , is there a way to re-export `compile-proto-file`?  It's an executable, not a library.  Also, would we have to choose new module names in order to enable re-export of library modules?  Maybe we could avoid that with package-qualified imports.

Gabriel439:
@j6carey: I don't think you would need to re-export executables like `compile-proto-file`.  At most, maybe just make the executable package a dependency of `proto3-suite` so that installing `proto3-suite` indirectly installs the executbale, but maybe even that is not necessary

For module names, you could use different names for the upstream packages and keep the original names for the `proto3-suite` package.  For example, for most of my packages I try to use the package name as the module name prefix.  So, for example, all of the `proto3-base` modules would have a `Proto3.Base` prefix

j6carey:
I still want to do something along these lines, but I'll have to restart the attempt.  Closing this PR.

PR #195, opened 2022-06-13
compile-proto-file: add --text-type option to generate code using strict Text

fumieval:
Context: https://github.com/awakesecurity/proto3-suite/pull/19

AFAIK, the use of Data.Text.Lazy is infrequent in the current ecosystem.
Adding an ability to toggle between strict and lazy texts releases us from tedious impedance matching

### DIFF
--- a/src/Proto3/Suite/DotProto/Generate.hs
+++ b/src/Proto3/Suite/DotProto/Generate.hs
@@ -22,6 +22,8 @@
 
 module Proto3.Suite.DotProto.Generate
   ( CompileError(..)
+  , StringType(..)
+  , parseStringType
   , TypeContext
   , CompileArgs(..)
   , compileDotProtoFile
@@ -40,6 +42,7 @@ import           Data.Coerce
 import           Data.Either                    (partitionEithers)
 import           Data.List                      (find, intercalate, nub, sortBy, stripPrefix)
 import qualified Data.List.NonEmpty             as NE
+import           Data.List.Split                (splitOn)
 import Data.List.NonEmpty (NonEmpty (..))
 import qualified Data.Map                       as M
 import           Data.Maybe                     (fromMaybe)
@@ -72,7 +75,15 @@ data CompileArgs = CompileArgs
   , extraInstanceFiles :: [FilePath]
   , inputProto         :: FilePath
   , outputDir          :: FilePath
+  , stringType         :: StringType
   }
+
+data StringType = StringType String String
+
+parseStringType :: String -> Either String StringType
+parseStringType str = case splitOn "." str of
+  xs@(_ : _ : _) -> Right $ StringType (intercalate "." $ init xs) (last xs)
+  _ -> Left "--string-type must be in the form of Module.Type"
 
 -- | Generate a Haskell module corresponding to a @.proto@ file
 compileDotProtoFile :: CompileArgs -> IO (Either CompileError ())
@@ -88,7 +99,7 @@ compileDotProtoFile CompileArgs{..} = runExceptT $ do
   Turtle.mktree (Turtle.directory modulePath)
 
   extraInstances <- foldMapM getExtraInstances extraInstanceFiles
-  haskellModule <- renderHsModuleForDotProto extraInstances dotProto importTypeContext
+  haskellModule <- renderHsModuleForDotProto stringType extraInstances dotProto importTypeContext
 
   liftIO (writeFile (Turtle.encodeString modulePath) haskellModule)
   where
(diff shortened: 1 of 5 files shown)


j6carey:
I've been working on a related change to support `ShortText` when desired, and one of the issues is that arose is the protobuf wrapper type `StringValue`: currently you have to match the Haskell string type used by that wrapper, or things will not build.  I have a solution for that, but I've been distracted by other issues.  How urgent is this change for you?

fumieval:
It's not urgent as we can stick to the fork for the time being. Supporting ShortText sounds useful - I should probably change the interface so that it would accept strict, lazy and short text types from the command line argument

fumieval:
@j6carey I updated the command line option so that you will be able to add `ShortText`.

j6carey:
@fumieval , there is still the matter of how to handle wrapper types.  I hope to be able to get to this soon.  Meanwhile, about the command line option: perhaps it would be better to have the option to the command be the qualified name of the type to use?  That would make it completely forward-compatible, even to text types we do not yet know about.  E.g.:

`--string-type Data.Text.Lazy.Lazy`

(Here I'm using "string" instead of "text" because "string" is the protobuf terminology, and that's what is a constant here; in principle the Haskell types could be called anything.)

fumieval:
That's a good point; updated the implementation

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -156,9 +167,10 @@ renameProtoFile filename =
 --   messages and enums.
 renderHsModuleForDotProto
     :: MonadError CompileError m
-    => ([HsImportDecl],[HsDecl]) -> DotProto -> TypeContext -> m String
-renderHsModuleForDotProto extraInstanceFiles dotProto importCtxt = do
-    haskellModule <- hsModuleForDotProto extraInstanceFiles dotProto importCtxt
+    => StringType
+    -> ([HsImportDecl],[HsDecl]) -> DotProto -> TypeContext -> m String
+renderHsModuleForDotProto custom extraInstanceFiles dotProto importCtxt = do
+    haskellModule <- hsModuleForDotProto custom extraInstanceFiles dotProto importCtxt
```
j6carey, line 173: The type of argument added here is `StringType`, not `ImportCustomisation`, so perhaps the new argument should be named `stringType` instead of `custom`?

There are also a few other places where the same issue arises.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -202,12 +217,19 @@ hsModuleForDotProto
 
        let hasService = has (traverse._DotProtoService) protoDefinitions
 
-       let importDeclarations = concat [ defaultImports hasService, extraImports, typeContextImports ]
+       let importDeclarations = concat
+              [ defaultImports
+                ImportCustomisation
+                  { usesGrpc = hasService
+                  , ..
```
j6carey, line 224: `RecordWildCards` are useful in many contexts, but perhaps they obscure things here?  I would prefer using `NamedFieldPuns` and mentioning the field to be copied from the argument by the same name.

src/Proto3/Suite/DotProto/Generate.hs
```diff
@@ -1787,8 +1820,13 @@ dpPrimTypeE ty =
         Float    -> wrap "Float"
         Double   -> wrap "Double"
 
-defaultImports :: Bool -> [HsImportDecl]
-defaultImports usesGrpc =
+data ImportCustomisation = ImportCustomisation
+  { strType :: StringType
```
j6carey, line 1824: To reduce the number of spellings to keep track of, I would prefer `stringType` over `strType`, to match the type name.

j6carey (changes requested):
Thanks for making `--string-type` more flexible.

In addition to the minor comments below, there are two more important issues:
1. The tests do not yet build; probably some of the new function arguments need to be supplied.
2. I did some manual testing and verified that, as expected, there are problems when using fields of type `google.protobuf.StringValue` unless the use of `--string-type` is the same between compilation of the Google wrapper .proto and compilation of the message making use of the `StringValue` wrapper, because of the way `StringValue` gets bypassed in the Haskell code.  I have an idea for how to solve this problem: #203 .

fumieval:
Thanks, I'm going to rebase the branch

src/Proto3/Suite/DotProto/Generate.hs
fumieval: fixed

src/Proto3/Suite/DotProto/Generate.hs
fumieval: There was a name conflict issue; I prefixed the field name instead

src/Proto3/Suite/DotProto/Generate.hs
fumieval: Sure

src/Proto3/Suite/DotProto/Generate.hs
j6carey: OK.

src/Proto3/Suite/DotProto/Generate.hs
j6carey: Thanks.

src/Proto3/Suite/DotProto/Generate.hs
j6carey: Thanks.

j6carey (approved):
The changes look good to me, and I see builds succeeding when I run them by hand with various Nix options.  Thanks for this new feature!

PR #196, opened 2022-06-15
Pin turtle dep to < 1.6.0 due to breaking FilePath change

morucci:


TristanCacqueray:
In the cabal.project file, we might be able to force the appropriate turtle version with https://cabal.readthedocs.io/en/3.4/cabal-project.html#solver-configuration-options

Gabriella439:
I published a matching revision to Hackage

PR #197, opened 2022-06-15
Update pinning of proto3-wire to < 1.4.0

morucci:
proto3-wire has been released recently and is available on hackage.


Gabriella439:
Thank you!

morucci:
Hi would you like to add a revision on hackage for that fix as well ?

Gabriella439:
Yeah, no problem!  I just published a matching revision

PR #199, opened 2022-06-17
Build against `turtle-1.6.0`

Gabriella439:
This builds against the older versions of `turtle`, too, albeit
at the expense of several deprecation warnings.  Once we cut a
release with this fix we can do a second pass to fix all the
warnings, that way end users have a smooth migration path.


Gabriella439:
One of the test failures is expected, but the other test failure caught a bug in `turtle-1.6.0`, which I'm fixing here: https://github.com/Gabriella439/turtle/pull/435

ixmatus (approved):

PR #203, opened 2022-07-01
Delegate to a polymorphic wrapper type for wrapped fields

j6carey:
...instead of delegating to code generated from the .proto
definitions of the standard Google protobuf wrapper types.

By avoiding delegation to wrapper implementations generated
from .proto definitions we avoid the need to generate such
implementations and link against them.

More importantly, by coercing between "Maybe Text" and
a polymorphic wrapper instead of a single fixed Haskell
representation of "StringValue", we enable a future
per-module choice between strict and lazy "Text" types,
rather than restricting ourselves to a uniform decision.
In fact, "Text" could be replaced by any string type that
provides the required type class instances for protobuf
serialization.  The same could be done for BytesValue.

### DIFF
--- a/tests/SimpleDecodeDotProto.hs
+++ b/tests/SimpleDecodeDotProto.hs
@@ -10,12 +10,14 @@ import Test.Tasty.HUnit (Assertion, (@?=), (@=?), testCase)
 import qualified Data.Map as M
 import Proto3.Suite
 import qualified Data.ByteString.Char8 as BC
+import GHC.Stack (HasCallStack)
 import System.IO
 
 import TestProto
 import qualified TestProtoImport
 import qualified TestProtoOneof
 import qualified TestProtoOneofImport
+import qualified TestProtoWrappers
 
 main :: IO ()
 main = do putStr "\n"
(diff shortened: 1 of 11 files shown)


tests/SimpleDecodeDotProto.hs
```diff
@@ -7,15 +7,44 @@ module Main where
...
+  maybeFormat <- lookupEnv "FORMAT"
+  case maybeFormat of
+    Nothing -> die $
+      "Please export a value for the FORMAT environment variable: one of: " ++
+      (intercalate ", " (map show ([minBound .. maxBound] :: [Format])))
+    Just formatStr -> case readEither formatStr of
+      Left msg -> die $
+        "Bad value of FORMAT environment variable: " ++ show format ++
+        ": " ++ show msg
+      Right f -> pure f
```
jcarr-awake, line 47: Since you're already using implicit parameters, I would define this in `main` instead, and pass it within the context of `tests` (Each of the terms becomes `(?format :: Format) => TestTree`).

Otherwise this gets a bit confusing to me, I had trouble understanding `testCaseInFormat`

j6carey: Changed by 03fd66aad1eb978f9fb589745d3afdd1cd5055d9; does that resolve the issue?

jcarr-awake (approved):
LGTM

PR #210, opened 2022-07-17
Eliminate a number of compiler warnings.

j6carey:
These all relate to Turtle functions deprecated by turtle-1.6.

Eliminating the warnings seems to make doctest behave more predictably.

We could use -Wno-deprecations, but this change is more selective.


jcarr-awake (approved):
Looks straightforwardly correct

Gabriella439:
Note that once you cut a release you can get rid of the `Turtle.Compat` module.  I only kept around the deprecated functions so that there would be at least one release that built against `turtle-1.5` and `turtle-1.6`

PR #214, opened 2022-07-25
Support parsing qualified option names.

riz0id:
Fixes the parser for protobuf option identifiers to support qualified option names such as:

```protobuf
syntax = "proto3";
import "foo";
option (foo.boolean_option_name_file) = true;
```


ixmatus (approved):
@riz0id I think we should add a test before you merge, to verify the parser behaves as expected. Otherwise, LTM.

PR #215, opened 2022-07-25
Fix tests of the Swagger schema of Google protobuf wrapper fields.

j6carey:
It seems that attempting to use conditional compilation on doctests
of the Swagger format for fields of Google protobuf wrapper type
silently suppresses those tests.

This commit replaces such doctests with HUnit tests.

### DIFF
--- a/tests/TestCodeGen.hs
+++ b/tests/TestCodeGen.hs
@@ -51,31 +51,31 @@ codeGenTests = testGroup "Code generator unit tests"
 
 swaggerWrapperFormat :: TestTree
 swaggerWrapperFormat = testGroup "Swagger Wrapper Format"
-    [ prop @TestProtoWrappers.TestDoubleValue
+    [ expectSchema @TestProtoWrappers.TestDoubleValue
            "{\"properties\":{\"wrapper\":{\"format\":\"DoubleValue\",\"type\":\"number\"}},\"type\":\"object\"}"
            "{\"properties\":{\"wrapper\":{\"format\":\"double\",\"type\":\"number\"}},\"type\":\"object\"}"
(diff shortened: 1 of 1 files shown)


tests/TestCodeGen.hs
```diff
@@ -46,6 +49,60 @@ codeGenTests = testGroup "Code generator unit tests"
...
+           "{\"properties\":{\"wrapper\":{\"format\":\"byte\",\"type\":\"string\"}},\"type\":\"object\"}"
+    ]
+  where
+    prop ::
+      forall a .
+      (ToSchema a, Typeable a) =>
+      LBS.ByteString ->
+      LBS.ByteString ->
+      TestTree
+    prop wrapperFormat noWrapperFormat =
```
jcarr-awake, line 89: What about a slightly clearer name for this, expectSchema? Or something else? If not for the context I have this test would be very unclear

j6carey: Good point; how does 02b0f881f3b3a39b09f6896d7d9cdd9760921748 look?

jcarr-awake: Well you didn't change the definition :)

jcarr-awake: But should be good in that case

j6carey: Hopefully fixed by f3e273a55ff770e47bd4f62b1a42171ae037f162 .

jcarr-awake (approved):

tests/TestCodeGen.hs
j6carey: CI passed.  Resolving this issue.

PR #217, opened 2022-07-25
Message extension parsing support.

riz0id:
This PR extends the proto3 parser to support parsing message extensions. Message extensions that are parsed are discarded for now (theres no way to represent message extensions in the current AST), but the parser will no longer produce a parse error "expecting enum" when `extend` is encountered.

### DIFF
--- a/src/Proto3/Suite/DotProto/Parsing.hs
+++ b/src/Proto3/Suite/DotProto/Parsing.hs
@@ -22,6 +22,10 @@ module Proto3.Suite.DotProto.Parsing
   , pFieldOptionStmt
   , pOptionId
   , pOptionKw
+
+    -- * Extension Parsers
+  , pExtendStmt
+  , pExtendKw
   ) where
 
 import Prelude hiding (fail)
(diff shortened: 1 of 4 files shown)


src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -195,19 +199,24 @@ sortStatements modulePath statements
...
-           <|> DPSPackage    <$> package
-           <|> DPSOption     <$> pOptionStmt
-           <|> DPSDefinition <$> definition
-           <|> DPSEmpty      <$  empty
+topStatement = 
+  choice 
+    [ DPSImport <$> import_
+    , DPSPackage <$> package
+    , DPSOption <$> pOptionStmt
+    , DPSEmpty <$ try pExtendStmt
```
ixmatus, line 216: Why do we need `try` here but no where else?

src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -225,9 +234,12 @@ package = do symbol "package"
              return $ DotProtoPackageSpec p
 
 definition :: ProtoParser DotProtoDefinition
-definition = message
-         <|> enum
-         <|> service
+definition = 
+  choice 
+    [ try message
```
ixmatus, line 239: Why do we need `try` added here? I believe `choice` is just a fold with `<|>` as its binary function so shouldn't `choice` with no `try` be equivalent to the prior formulation?

src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -412,4 +424,22 @@ reservedField :: ProtoParser [DotProtoReservedField]
...
+-- Message Extensions ----------------------------------------------------------
+
+pExtendStmt :: ProtoParser (DotProtoIdentifier, [DotProtoMessagePart])
+pExtendStmt = do 
+  pExtendKw
+  idt <- singleIdentifier
+  fxs <- braces (many messagePart)
+  pure (idt, fxs)
+
+-- | Parses a single keyword token "option".
```
ixmatus, line 438:
```suggestion
-- | Parses a single keyword token "extend".
```

ixmatus (approved):
LGTM pending some comments regarding use of `try`.

src/Proto3/Suite/DotProto/Parsing.hs
riz0id: The choice between `extend` (on line 215) and `enum` requires backtracking since lookahead is greater than 1.

src/Proto3/Suite/DotProto/Parsing.hs
riz0id: Ideally I could just have written `try pExtendStmt <|> enum` within the `definition` parser, but because theres no AST to represent message extensions I need to parse extensions at the top-level return `DPSEmpty` then bubble up backtracking the between `enum` and `extend` to this parser.

ixmatus: Okay, thanks for explaining!

PR #218, opened 2022-07-27
Fix Import Render

tdbgamer:
It looks like if you parse and then render a proto file, it strips the double quotes from the imports accidentally. This fixed it for me, so just wanted to contribute it back.

PR #220, opened 2022-07-29
Fix parsing and pretty-printing of string literals

Gabriella439:
While debugging the test failure on #218 I realized that the
logic for parsing and pretty-printing import paths was broken in
several ways:

* Import paths were being rendered unquoted (so the rendered .`proto`
  was not valid)

* Import parsing and pretty-printing was not handling escape sequences
  correctly

* The quickcheck generator was not even exercising this code path

  … which is why they remained broken

… so I fixed this

Fixes #218


src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -125,9 +126,71 @@ globalIdentifier = token $ string "." >> _identifier
 ----------------------------------------
 -- values
 
--- [issue] these string parsers are weak to \" and \000 octal codes
-stringLit :: ProtoParser String
-stringLit = stringLiteral <|> stringLiteral'
+strLit :: ProtoParser String
+strLit = doubleQuotedLiteral <|> singleQuotedLiteral
```
j6carey, line 130: To understand the correctness of this parser I would need to find the part of the proto3 specification that specifies the grammar for string literals look like in .proto files, but thus far I have been unable to locate that specification.  Do you know where that documentation resides?

Gabriella439: You can find the grammar here: https://developers.google.com/protocol-buffers/docs/reference/proto3-spec

The relevant part is:

```abnf
strLit = ( "'" { charValue } "'" ) |  ( '"' { charValue } '"' )
charValue = hexEscape | octEscape | charEscape | /[^\0\n\\]/
hexEscape = '\' ( "x" | "X" ) hexDigit hexDigit
octEscape = '\' octalDigit octalDigit octalDigit
charEscape = '\' ( "a" | "b" | "f" | "n" | "r" | "t" | "v" | '\' | "'" | '"' )
```

j6carey: Thank you very much!

j6carey (approved):
Thank you very much for this very thorough fix.

src/Proto3/Suite/DotProto/Parsing.hs
Gabriella439: You're welcome!

Gabriella439:
I had to remove a few tests which were broken.  The root cause was that the proto3 parser was silently ignoring the rest of the file when hitting an invalid parse instead of failing.  This PR fixed that (by requiring the parser to match `eof` when it's done), but then unearthed some invalid proto3 files in the test suite, which I had to delete (along with their reverse dependencies).

j6carey:
Thanks!

Gabriella439:
You're welcome!

PR #221, opened 2022-08-12
Fix build issues

fumieval:
* Enabled FlexibleContexts in Proto3/Suite/DotProto/Generate/Swagger/Wrappers.hs
* Fixed generated code for `Qualified` DotProtoIdentifier


Gabriella439:
Thank you for fixing that!

PR #226, opened 2022-12-07
Support --string-type Data.Text.Short.ShortText

j6carey:
Also replace newtype OverrideToSchema with String and Bytes newtypes in
hopes of making it easier to understand how type class instances are selected.


src/Proto3/Suite/DotProto/Generate/Syntax.hs
```diff
@@ -20,7 +21,14 @@ haskellNS = Module "Hs"
 --
 
 apply :: HsExp -> [HsExp] -> HsExp
-apply f = HsParen . foldl HsApp f
+apply f = paren . foldl HsApp f
+
+applyMaybe :: Maybe HsExp -> HsExp -> HsExp
+applyMaybe = maybe id (\f x -> paren (HsApp f (paren x)))
```
jcarr-awake, line 27: It seems like all the uses of applyMaybe are flipped, so maybe this could be changed, but I'm not sure if there is a good name for that...

tests/TestCodeGen.hs
```diff
@@ -41,18 +41,20 @@ codeGenTests = testGroup "Code generator unit tests"
   , camelCaseMessageFieldNames
   , don'tAlterEnumFieldNames
   , knownTypeMessages
+  , pythonInteroperation
+  ]
+
+pythonInteroperation :: TestTree
+pythonInteroperation = testGroup "Python interoperation" $ do
```
jcarr-awake, line 48: I'm not sure I understand the connection between this test addition and the PR as a whole. Is there some connection or was it just something that happened to be roped in?

tests/TestCodeGen.hs
```diff
@@ -41,18 +41,20 @@ codeGenTests = testGroup "Code generator unit tests"
...
-  , simpleDecodeDotProto RegularRecords "Binary"
-  , simpleEncodeDotProto RegularRecords "Jsonpb"
-  , simpleDecodeDotProto RegularRecords "Jsonpb"
+  recStyle <- [RegularRecords]
 #endif
-  ]
+  tt <- ["Data.Text.Lazy.Text", "Data.Text.Text", "Data.Text.Short.ShortText"]
+  format <- ["Binary", "Jsonpb"]
+  direction <- [simpleEncodeDotProto, simpleDecodeDotProto]
+  [direction recStyle tt format]
```
jcarr-awake, line 57: I would prefer using pure vs. list syntax. It's a minor difference but it fits into the "schema" of do blocks more than the list syntax does if that makes sense.

jcarr-awake (commented):
This seems to all make sense although there might be some changes entangled that make it more confusing. We should ensure all of them are documented in the changelog.

tests/TestCodeGen.hs
j6carey: Before we were failing to test all the different varieties of string representation.  I want to test all 3: `Data.Text.Lazy.Text`, `Data.Text.Text`, `Data.Text.Short.ShortText`.  Therefore I'm multiplying the number of tests by 3, and it is going to get verbose and error-prone to write out all those combinations by hand.  That's why I'm generating the list of tests now.

tests/TestCodeGen.hs
j6carey: Here I was thinking it would make it clear which `Monad` is in play here.  But OK.

j6carey: Changed to `pure @[]` in babd2a8195abb03f2b9180da83d15410eea7fcc3.

src/Proto3/Suite/DotProto/Generate/Syntax.hs
j6carey: What do you think of `maybeModify`?: babd2a8195abb03f2b9180da83d15410eea7fcc3

tests/TestCodeGen.hs
j6carey: Oh, I should mention that when large records are supported, we used to test only them, but now we also test regular records in that case.

tests/TestCodeGen.hs
jcarr-awake: That's a good choice

j6carey:
@jcarr-awake , does 889e5edc22089674415c3d43c0d31221ad71d5ab help with your concerns over `CHANGELOG.md`?

jcarr-awake (approved):

PR #228, opened 2023-02-03
Add `--no-field-prefix` option

ccycle:
Added option to remove field name prefixes from records generated by `compile-proto-file`. This change is intended for use with `NoFieldSelectors` for auto-generated records.

### DIFF
--- a/tools/compile-proto-file/Main.hs
+++ b/tools/compile-proto-file/Main.hs
@@ -49,7 +49,7 @@ parseArgs = info (helper <*> parser) (fullDesc <> progDesc "Compiles a .proto fi
 
     isPrefixed = IsPrefixed . not <$> switch (
       long "no-field-prefix"
-      <> help "Remove prefix for record field names"
+      <> help "Do not prefix type names to record field names"
       )
 
 main :: IO ()


src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -508,13 +512,34 @@ prefixedMethodName serviceName (x : xs)
...
   field <- typeLikeName fieldName
   return (fieldLikeName msgName ++ field)
 
+prefixedFieldNameWithFlag :: MonadError CompileError m => IsPrefixed -> String -> String -> m String
+prefixedFieldNameWithFlag (IsPrefixed flag) msgName fieldName = do
+  if flag then prefixedFieldName msgName fieldName else return $ if name `elem` keywords then name ++ "_" else toCamelCase fieldName
+  where
+    name = fieldLikeName fieldName
+    -- copy from https://hackage.haskell.org/package/hscolour-1.20.3/docs/src/Language-Haskell-HsColour-Classify.html#keywords
+    -- and remove "forall", "qualified", "ccall", "as", "safe", "unsafe"
```
j6carey, line 535: Could you mention why those keywords are removed from this check?

Also, I would rather not copy the list because there is a license mismatch.  How about an equivalent predicate, such as a case statement on the first letter, followed by `elem` on a smaller list of keywords with that first letter?  That would be faster anyway.

tests/TestCodeGen.hs
```diff
@@ -51,10 +51,11 @@ pythonInteroperation = testGroup "Python interoperation" $ do
 #else
   recStyle <- [RegularRecords]
 #endif
+  isPrefixedArg <- [IsPrefixed True, IsPrefixed False]
```
j6carey, line 54: Comment only: this will make these tests take twice as long, and they are already lengthy, but I think it is still acceptable.  If that changes, we can consider random exploration of the configuration space.

src/Proto3/Suite/DotProto/Internal.hs
```diff
@@ -508,13 +512,34 @@ prefixedMethodName serviceName (x : xs)
       method <- typeLikeName (x : xs)
       return (fieldLikeName serviceName ++ method)
 
+prefixedMethodNameWithFlag :: MonadError CompileError m => IsPrefixed -> String -> String -> m String
+prefixedMethodNameWithFlag _ _ "" = invalidTypeNameError "<empty name>"
+prefixedMethodNameWithFlag (IsPrefixed flag) serviceName (x : xs)
+  | flag = prefixedMethodName serviceName (x : xs)
+  | isLower x = return (fieldLikeName (x : xs))
+  | otherwise = fieldLikeName <$> typeLikeName (x : xs)
+
```
j6carey, line 521: Do you need a keyword check in this function too?

j6carey (changes requested):
Thanks for this feature.  Mostly this looks good, though I do have some requests.

src/Proto3/Suite/DotProto/Internal.hs
j6carey: Yes, but if you want to go with a Set that's fine too.

But to clarify my first question: why are you allowing "forall", "qualified", "ccall", "as", "safe", and "unsafe"?

src/Proto3/Suite/DotProto/Internal.hs
j6carey: OK, so if it is needed, then could you add it?

j6carey: Does `prefixedMethodNameWithFlag` also need a keyword check?

src/Proto3/Suite/DotProto/Internal.hs
j6carey: Thanks.

src/Proto3/Suite/DotProto/Internal.hs
riz0id: I strongly suggest not trying to do any case convention changes and just use the name of service methods and message fields that are provided in the protobuf source. I can't think of any situations where there is a benefit to modifying the lexical content names. Also, theres no function that exists that can convert snake case to camel case without opening up the potential for unintended name shadowing in the code generator.

If some names are colliding with Haskell keywords I think the best solution is to add a `hs_name` protobuf option that can override the default name the code generator chooses. Many languages already support a similar option, for example Java's `outer_classname`.

riz0id: I also strongly suggest not making tying Haskell's grammar to the code generators internals. If have special naming conditions when a protobuf identifier is using a Haskell keyword and a new keyword is added to Haskell, then we are forced to add that keyword as another special case. This could cause backwards compatibility issues when people update to a new version `proto3-suite` and see that a `"_" :: String` is now being appended to the end of generated record names.

I really feel like the best attitude towards names is none at all. We should only use names as they appear in the protobuf source with no modifications, but also add fine grained control that allows users to override names in special situations.

riz0id: @ccycle I spoke with @j6carey about your changes. We agreed that the keyword renaming rule should be avoided unless you absolutely need it for your particular use case, but that supporting a way to override names would be the best way to handle exceptional cases. Theres quite a bit of work needed to add support options into the `proto3-suite` AST so the `hs_name` solution probably isn't feasible right now. If you do need protobuf names to avoid colliding with keywords then the changes you have here are fine for now, but will probably be removed at a later point when we support `hs_name`.

riz0id: In this PR and until we have a `hs_name` option, we should stick to the naming convention rules that the rest of `proto3-suite` uses for record selector names which is what you do here with `toCamelCase`. In the future though, when we have facilities to override names in generated Haskell code, we will probably make breaking changes to instead use snake case for records. Currently the number of edge cases that need to be considered in order to rename protobuf names to camel case Haskell identifiers is getting out of hand. Using snake case record selectors makes a lot more sense (imo) since:

1. Protobuf suggests snake case for message field names by convention.
2. Snake case would not require specific rules to avoid collisions with keywords or double underscores (i.e. "__").
3. Many people (myself included) already use snake case by convention to distinguish record selectors from ordinary functions.

tldr; Yes use `toCamelCase` for now, but once `proto3-suite` supports a `hs_name` option expect breaking changes that use snake case for message field names.

j6carey (approved):
My concerns have been addressed, but @riz0id , did you want the keyword checks eliminated, or are they OK until `hs_name` is available?

src/Proto3/Suite/DotProto/Internal.hs
j6carey: My concerns have been addressed, but @riz0id , did you want the keyword checks eliminated, or are they OK until `hs_name` is available?

tools/compile-proto-file/Main.hs
```diff
@@ -47,5 +47,10 @@ parseArgs = info (helper <*> parser) (fullDesc <> progDesc "Compiles a .proto fi
       $ long "largeRecords"
       <> help "Use large-records library to optimize the core code size of generated records"
 
+    isPrefixed = IsPrefixed . not <$> switch (
+      long "no-field-prefix"
+      <> help "Remove prefix for record field names"
```
fumieval, line 52: This is a nitpicking, but perhaps more natural explanation of the behaviour would be "do not prefix record field names by the type names", since it does not remove anything when this is enabled.

src/Proto3/Suite/DotProto/Internal.hs
fumieval: @riz0id Do you have any remaining concerns?

PR #239, opened 2023-08-06
Fix compilation error when swagger is disabled

TristanCacqueray:
When building without the swagger flag, this change fixes the following error:

  src/Proto3/Suite/DotProto/Generate.hs:792:46: error: [GHC-39999]
      • Could not deduce ‘Functor f0’ arising from a use of ‘<$>’
        from the context: MonadError CompileError m


j6carey (approved):
Looks good to me.  Thanks!

PR #244, opened 2023-09-26
Nix: if enableLargeRecords is false use -f-large-records

j6carey:
The absence of "-flarge-records" is not enough because the "large-records" cabal flag is enabled by default.

This fix should improve the coverage of continuous integration tests for the proto3-suite repository.

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -50,7 +50,7 @@ in {
                 (if enableDhall then "-fdhall" else "")
                 (if enableSwagger then "" else "-f-swagger")
                 (if swaggerWrapperFormat then "-fswagger-wrapper-format" else "")
-                (if enableLargeRecords then "-flarge-records" else "")
+                (if enableLargeRecords then "" else "-f-large-records")
               ];
             in
             (haskellPackagesNew.callCabal2nixWithOptions
@@ -65,7 +65,7 @@ in {
                 ++ (if enableDhall then [ "-fdhall" ] else [ ])
                 ++ (if enableSwagger then [ "" ] else [ "-f-swagger" ])
                 ++ (if swaggerWrapperFormat then [ "-fswagger-wrapper-format" ] else [ "" ])
-                ++ (if enableLargeRecords then [ "-flarge-records" ] else [ ]);
+                ++ (if enableLargeRecords then [ ] else [ "-f-large-records" ]);
             });
 
           proto3-suite-boot =


riz0id (changes requested):
This is just a problem with how the flag is defined though, no? Currently theres no way to disable large records, the default value for the flag needs to be changed from `True` to `False`.
```
flag large-records
  Description:   Generate records with smaller core code size using the large-records library
  Default:       True
  Manual:        True
```

j6carey:
> This is just a problem with how the flag is defined though, no? Currently theres no way to disable large records, the default value for the flag needs to be changed from `True` to `False`.

I'm able to disable the flag with `-f-large-records`; that's what this PR does.

riz0id:
> > This is just a problem with how the flag is defined though, no? Currently theres no way to disable large records, the default value for the flag needs to be changed from `True` to `False`.
> 
> I'm able to disable the flag with `-f-large-records`; that's what this PR does.

I don't understand why passing `-f-large-records` should disable large records, seems like `-f-large-records` should _enable_ large records. In any case, I don't understand how passing the `-f-large-records` flag is disabling the `large-records` flag in the cabal if both the default and manual values for `large-records` is `True`.

riz0id (approved):

riz0id:
If it does work then I'll approve it. It seems to me like value for `Manual` should be changed from `False` to `True`, the nix change you made here should be reverted. Or, the flag should be renamed to something like `disable-large-records` to clearly indicate that passing it to cabal disable compilation with large records.

j6carey:
> I don't understand why passing `-f-large-records` should disable large records, seems like `-f-large-records` should _enable_ large records. In any case, I don't understand how passing the `-f-large-records` flag is disabling the `large-records` flag in the cabal if both the default and manual values for `large-records` is `True`.

What I am seeing is that when you pass `-fx` on the `cabal` command line, `cabal` tries to enable `x`; whereas if you pass `-f-x` then it tries to disable `x`; otherwise it takes the default value from the `.cabal` file.

j6carey:
> If it does work then I'll approve it. It seems to me like value for `Manual` should be changed from `False` to `True`, the nix change you made here should be reverted. Or, the flag should be renamed to something like `disable-large-records` to clearly indicate that passing it to cabal disable compilation with large records.

The "manual" flag actually means something other than the meaning of the command-line argument for the flag when present:
https://cabal.readthedocs.io/en/stable/cabal-package.html#configuration-flags

> By default, Cabal will first try to satisfy dependencies with the default flag value and then, if that is not possible, with the negated value. However, if the flag is manual, then the default value (which can be overridden by commandline flags) will be used.

PR #246, opened 2023-10-06
Encode field numbers at compile time

j6carey:
...at the cost of some increased code size.

### DIFF
--- a/nix/overlays/haskell-packages.nix
+++ b/nix/overlays/haskell-packages.nix
@@ -39,8 +39,8 @@ in {
               source = pkgsNew.fetchFromGitHub {
                 owner = "awakesecurity";
                 repo = "proto3-wire";
-                rev = "6dcc557fc9d3ffcbfe45c6bea4969830f9e0de9c";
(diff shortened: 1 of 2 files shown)


ixmatus (approved):

riz0id:
@j6carey how do the inline pragmas benefit these instances?

j6carey:
> @j6carey how do the inline pragmas benefit these instances?

@riz0id , Speed has increased as a result of these changes, but I'm hoping to make some further refinements to this PR to reduce the code size increase, which I have also measured.  The theory behind the speed increase is to get the encodings of the field numbers to be computed at compile time, and to reduce the number of non-inline function calls and their associated overhead (saving to the stack and then restoring).

riz0id (approved):

jcarr-awake (approved):
Looks good pending good benchmark results

jcarr-awake:
Thanks for working to reduce the code size impact!

j6carey:
The benchmark results are good--actually better than what I measured before, though they are relative to a different baseline, and that may be the reason.

PR #247, opened 2024-01-25
Update Tutorial.hs, giving hint on first enum value

### LINKED ISSUE #235: Tutorial should provide more guidance around idiomatic enum use
In proto3 (and generally) it is recemmended to have an UNKNOWN enum value with value 0. The reason is, if the enum is missing or contains unknown value, the user will observe the default value 0 (**big note: this is only true for closed enums, which proto3-suite doesn't implement it seems. See later below**). If that is not UNKNOWN but some other specific value, the code interpreting the message might mistake the actually missing enum to an enum with the first possible value.

The tutorial in
https://github.com/awakesecurity/proto3-suite/blob/1f2c156b1178599d3853dac941beb3f29e2bdf5e/src/Proto3/Suite/Tutorial.hs#L122
defines `Circle = 0` which seems dangerous.

Sidenote (not sure why, but I wrote it): since proto3-suite's `Enumerated a` type seems to be a newtype for `Either Int32 a`, I assume the left value is an int that was unknown at parsing time. So, when I'm reading protobuf from Haskell code, I'm kind of forced to double-check the value - is it Right? Is it UNKNOWN? Well, that's just a small annoyance, and I think is good in spirit of proto intention (that is, as long as I don't care to inspect the field, I can reencode it and pass it along to other systems just fine, with the content unchanged, in the hope they can deal with the value if they need to).

So bottom line is, using non-UNKNOWN enum value as zero would kind of surely cause problem in other languages eventually.

Sadly, this forces us to add an extra not-really-valid constructor to the Haskell enum, making its nice pure use bit more painful (error checking creeps in).

I wonder if one could eat the cake and have it too - have some option for the codegen, that:
1) enum constructor with value = 0 is omitted from the Haskell-generated type, and
2) a value=0 is parsed into a `Left int` value (maybe that would happen after 1) automatically).

Wondering if there's any downside / semantic difference of that? Why don't other languages do that?

Actually take above with grain (or, hm, shovel) of salt: https://protobuf.dev/programming-guides/enum/ defines the madness of interpretation across proto2/3. Oh my. Based on that, it seems proto3's Enumerated type, when interpreted directly, implements the "open" enum. And for that usecase, what I said about needing to have UNKNOWN = 0 sounds not valid (the open enum avoids the default-returning problem of "closed" enums).

(Don't ask if the closed-enum case is well-supported by proto3-suite then... I also don't care much about that case since I'm pure-proto3, but others might do).

Recap:
- if you use open proto

So a bit of clarification of the documentation around this would be helpful.

Slightly related #28 .
Sorry for the braindump format.

robinp:
Tries to condense best practices for protobuf enum use. Fixes #235, kind of.

### DIFF
--- a/src/Proto3/Suite/Tutorial.hs
+++ b/src/Proto3/Suite/Tutorial.hs
@@ -103,6 +103,17 @@ instance Named Bar
 -- >   | Square
 -- >   | Triangle
 -- >   deriving (Bounded, Eq, Enum, Finite, Generic, Named, Ord, ProtoEnum)
+--
+-- Note: Due to how proto3 enums work, it is strongly advised that the first
+-- enum constructor is called Unknown (or something of similar meaning). If a
+-- proto3 wire format is missing the enum field, then the Haskell record
+-- resulting from parsing the wire format will still contain the
+-- zero-numbered enum constructor as present. Adding the Unknown constructor
+-- is especially important for future-proofing data, since if the enum
+-- is later added to a message as a field, then re-reading the previously
+-- persisted data will seem as if the enum value is present with that
+-- first value. If that first value is something rather specific, it can
+-- mislead the reader of that data.
 data Shape
   = Circle
   | Square


robinp:
Thank you for the feedback. I respect the opinion. May Github preserve the notes in our memories.

PR #248, opened 2024-05-02
Use "ghc" library in place of "haskell-src".

j6carey:
The "ghc" library is now used to parse and print Haskell source code.
Switching to "ghc" adds support for language features beyond Haskell 98
and improves diagnostic messages for sources specified with
"--extraInstanceFile".  Breakage should be limited to users of:

* Proto3.Suite.DotProto.Generate
* Proto3.Suite.DotProto.Generate.LargeRecord
* Proto3.Suite.DotProto.Generate.Syntax

To reflect the potential for breakage
we increase the library version to 0.8.0.

This change also replaces testing of
GHC 8.10.7 with testing of GHC 9.6.2
and GHC 9.8.1, and indicates in
CHANGELOG.md that we are dropping
support for GHC 8.10 and adding
support for GHC 9.6.

This change also fixes some build failures that
occurred when Swagger support is disabled,
and expands the flag combinations to all the
ones currently supported.

Support for dhall and large-records is limited
in more recent versions of GHC due to
problems with building those dependencies.


riz0id (approved):
LGTM. Thank you!

riz0id (approved):

PR #255, opened 2024-09-23
Remove unnecessary `DotProtoEmptyField` constructor

riz0id:
Minor change improving the AST by removing the `DotProtoEmptyField` constructor from the `DotProtoField`. This constructor complicated the AST and was needed nowhere.


j6carey (approved):
This looks good, but it may be that `DotProtoEmptyField` was an (apparently unsuccessful) attempt to handle the [empty statement](https://protobuf.dev/reference/protobuf/proto3-spec/#emptystatement).  Perhaps another PR could be created to add support for those, probably by making certain parser return `Maybe` types and the doing a `catMaybes` on the resulting list.

Issue #257, opened 2024-11-05
Parsing support for empty statements


j6carey:
See https://protobuf.dev/reference/protobuf/proto3-spec/#emptystatement

As far as we know this issue is low priority. If you encounter a need for empty statement support, please note it in this issue.

j6carey:
@riz0id  fixed this in #260; thanks!

PR #260, opened 2024-11-13
Support for empty statement

riz0id:
This PR adds support for parsing and discarding empty statements in enumerations, messages, and services. As well, this PR removes the "empty" constructors in `DotProtoServicePart` and `DotProtoMessagePart` as they're no longer used. This PR resolves issue #257.

### DIFF
--- a/src/Proto3/Suite/DotProto/Parsing.hs
+++ b/src/Proto3/Suite/DotProto/Parsing.hs
@@ -383,10 +383,11 @@ pOptionKw = do
 --------------------------------------------------------------------------------
 -- service statements
 
-servicePart :: ProtoParser DotProtoServicePart
-servicePart = DotProtoServiceRPCMethod <$> rpc
-          <|> DotProtoServiceOption <$> pOptionStmt
-          <|> DotProtoServiceEmpty <$ empty
+servicePart :: ProtoParser (Maybe DotProtoServicePart)
+servicePart =
+  (fmap (Just . DotProtoServiceRPCMethod) rpc)
+    <|> (fmap (Just . DotProtoServiceOption) pOptionStmt)
+    <|> Nothing <$ pEmptyStmt
 
 rpcOptions :: ProtoParser [DotProtoOption]
 rpcOptions = braces $ many pOptionStmt
@@ -407,10 +408,13 @@ rpc = do symbol "rpc"
          return RPCMethod{..}
 
 service :: ProtoParser DotProtoDefinition
-service = do symbol "service"
-             name <- singleIdentifier
-             statements <- braces (many servicePart)
-             return $ DotProtoService mempty name statements
+service = do
+  symbol "service"
+  name <- singleIdentifier
+  statements <- braces do
+    results <- many servicePart
+    pure (catMaybes results)
+  return $ DotProtoService mempty name statements
 
 --------------------------------------------------------------------------------
 -- message definitions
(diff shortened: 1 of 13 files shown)


src/Proto3/Suite/DotProto/Parsing.hs
```diff
@@ -378,10 +383,11 @@ pOptionKw = do
...
 -- service statements
 
-servicePart :: ProtoParser DotProtoServicePart
-servicePart = DotProtoServiceRPCMethod <$> rpc
-          <|> DotProtoServiceOption <$> pOptionStmt
-          <|> DotProtoServiceEmpty <$ empty
+servicePart :: ProtoParser (Maybe DotProtoServicePart)
+servicePart =
+  (fmap (Just . DotProtoServiceRPCMethod) rpc)
+    <|> (fmap (Just . DotProtoServiceOption) pOptionStmt)
```
j6carey, line 389: Presumably you used them for stylistic reasons, but the parentheses around the arguments to `<|>` are unnecessary.

test-files/test_proto_empty_field.proto
```diff
@@ -0,0 +1,15 @@
+
+syntax = "proto3";
+package TestProtoEmptyFeild;
```
j6carey, line 3: "Feild" looks misspelled.

tests/TestCodeGen.hs
```diff
@@ -254,6 +254,7 @@ compileTestDotProtos logger recStyle decodedStringType = do
   let protoFiles :: [Turtle.FilePath]
       protoFiles =
         [ "test_proto.proto"
+        , "test_proto_empty_field.proto"
```
j6carey, line 257: This will trigger code generation, and perhaps that's all you care about: does it parse without error?  But if you wanted to actually test the generated code's interoperability with generated Python code then you would have to also modify:
  * `tests/SimpleDecodeDotProto.hs`
  * `tests/SimpleEncodeDotProto.hs`
  * `tests/check_simple_dot_proto.py`
  * `tests/send_simple_dot_proto.py`

But for something like empty statements I would think you could do a much simpler check: parse the .proto file into AST and then just check that parsing is successful and that the AST contains message types with the expected fields, both before and after the empty statement.

j6carey (changes requested):
In addition to the specifics: are you planning to keep `DPSEmpty`?  Having an explicit representation of an empty top-level statement shows a different style than the approach you are taking with this PR in service, message, and enum definitions.

src/Proto3/Suite/DotProto/Parsing.hs
riz0id: Yes, we probably do. If we don't need `try` here then adding it wouldn't hurt.

test-files/test_proto_empty_field.proto
j6carey: Thanks.

src/Proto3/Suite/DotProto/Parsing.hs
j6carey: Thanks.

tests/TestCodeGen.hs
j6carey: Thanks for the revised tests.

j6carey (approved):
Thanks for the improvement.  I still wonder if we should do something about `DPSEmpty` being handled differently, but I don't want to block this PR for that.

riz0id:
> DPSEmpty



> Thanks for the improvement. I still wonder if we should do something about `DPSEmpty` being handled differently, but I don't want to block this PR for that.

Typically the top-level needs to be handled differently. I think leaving it for now is fine, but if it can be removed then I think we should remove it in the future.

PR #269, opened 2025-03-20
Add access to octets in MessageEncoding.

j6carey:
Also add `Eq` and `Show` instances for `MessageEncoding`,
rename `cacheMessageEncoder` to `cacheMessageEncoding`,
add `cacheMessage`, and rename `cachedPrefix` to `cacheFields`.

### DIFF
--- a/src/Proto3/Suite/Form/Encode/Core.hs
+++ b/src/Proto3/Suite/Form/Encode/Core.hs
@@ -30,16 +30,8 @@ module Proto3.Suite.Form.Encode.Core
   ( MessageEncoder(..)
   , toLazyByteString
   , etaMessageEncoder
-  , MessageEncoding
-  , cacheMessageEncoding
-  , cachedMessageEncoding
-  , messageEncodingToByteString
-  , unsafeByteStringToMessageEncoding
   , Prefix(..)
   , etaPrefix
-  , Fields
-  , cachePrefix
-  , cachedFields
   , Distinct
   , DistinctCheck
   , RepeatedNames
@@ -53,8 +45,9 @@ module Proto3.Suite.Form.Encode.Core
   , Occupy1
   , NameSublist
   , omitted
+  , SFieldNumberI
   , KnownFieldNumber
-  , fieldNumber
+  , fieldNumberVal
   , Field(..)
   , FieldForm(..)
   , PackedFieldForm(..)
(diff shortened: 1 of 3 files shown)


src/Proto3/Suite/Form/Encode/Core.hs
```diff
@@ -96,34 +98,56 @@ toLazyByteString = Encode.toLazyByteString . untypedMessageEncoder
...
 
 type role MessageEncoding nominal
 
+instance (Message message, Show message) =>
+         Show (MessageEncoding message)
+  where
+    showsPrec d (messageEncodingToByteString -> bs) = showParen (d >= 11) $
+      case fromByteString bs of
+        Left _ ->
+          showString "Proto3.Suite.Form.Encode.unsafeByteStringToMessageEncoding " .
```
riz0id, line 116: Use `show 'unsafeByteStringToMessageEncoding` (name quoting) to ensure that this qualified name stays up to date, and to ensure we get a compile-time error should the name of `unsafeByteStringToMessageEncoding` change.

tests/Main.hs
```diff
@@ -363,6 +364,20 @@ encodeCachedSubmessage = testGroup "Cached Submessages"
...
+    testShow :: TestTree
+    testShow = testCase "Show MessageEncoding" $ do
+      let trivial = TP.Trivial 123
+          badForm = "abc"
+          encoding :: FormE.MessageEncoding TP.Trivial
+          encoding = cacheReflection trivial
+          bogus :: FormE.MessageEncoding TP.Trivial
+          bogus = FormE.unsafeByteStringToMessageEncoding badForm
+      -- Test desired behavior for valid encoding:
+      show encoding @?= "Proto3.Suite.Form.Encode.messageReflection " ++ showsPrec 11 trivial ""
```
riz0id, line 376: Use `'messageReflection`

tests/Main.hs
```diff
@@ -363,6 +364,20 @@ encodeCachedSubmessage = testGroup "Cached Submessages"
...
+      -- Test fallback behavior for invalid encoding:
+      show bogus @?=
+        "Proto3.Suite.Form.Encode.unsafeByteStringToMessageEncoding " ++ showsPrec 11 badForm ""
```
riz0id, line 379: Use `'unsafeByteStringToMessageEncoding`

riz0id (approved):

src/Proto3/Suite/Form/Encode/Core.hs
j6carey: Neat idea!  Implemented in ecd3d03a1d5c9030133c5db05eb3a5da2dbe634f .

tests/Main.hs
j6carey: Actually, I kept the string literals in the test because otherwise we might not notice if the module qualification changing from an exposed module to a module that is not exposed by the library, such as `Proto3.Suite.Form.Encode.Core`.

tests/Main.hs
j6carey: Actually, I kept the string literals in the test because otherwise we might not notice if the module qualification changing from an exposed module to a module that is not exposed by the library, such as `Proto3.Suite.Form.Encode.Core`.

PR #274, opened 2025-04-10
Support aeson-2.2.

j6carey:
This initial support for aeson-2.2 does not make
use of its new features around default values.

This change also sets the default GHC version
in the nix-shell environment to 9.10.1 to avoid
problems with building cabal-install.


shell.nix
```diff
@@ -7,7 +7,7 @@
 #   $ nix-shell --arg fast true
 #
 { fast ? false
-, compiler ? "ghc948"
+, compiler ? "ghc9101"
```
riz0id, line 10: How come we are updating this?

riz0id (approved):

shell.nix
j6carey: The nix-shell environment doesn't work for 9.4.8.  Some problem with the `unix` package, IIRC.

riz0id: Gotcha, I'm looking into that now. Hence, #275.

PR #275, opened 2025-04-10
Reduce number of overridden Nix packages

riz0id:
This PR attempts to cut back on the number of dependencies that require overrides to their derivation before those packages can be used.

- Removed [`range-set-list`](https://hackage.haskell.org/package/range-set-list): this package was only used to normalize collections of integer intervals in the `canonicalize-proto-file` executable. This package needed to be jail broken in order to be used, and has caused a number of package conflict issues for myself and others in the past.


tools/canonicalize-proto-file/Main.hs
```diff
@@ -30,9 +30,54 @@ data Args w = Args
...
   , proto      :: w ::: FilePath   <?> "Path to input .proto file"
   } deriving Generic
+
 instance ParseRecord (Args Wrapped)
+
 deriving instance Show (Args Unwrapped)
 
+-- | @('isOverlappingIntervals' a b :: 'Bool')@ is a relation between two 
+-- interval-like types @a@ and @b@ that is 'True' when the two intervals are 
+-- overlapping or touching over a subinterval of @a@ and @b@.
```
j6carey, line 40: This comment does not match the type signature.

tools/canonicalize-proto-file/Main.hs
```diff
@@ -30,9 +30,54 @@ data Args w = Args
...
+isOverlappingIntervals :: (Ord a, Enum a) => (a, a) -> (a, a) -> Bool
+isOverlappingIntervals (x, y) (u, v) = y >= u || v <= x
```
j6carey, line 42: Is this a problem, or is there an extra precondition somewhere?
```
ghci> let isOverlappingIntervals (x, y) (u, v) = y >= u || v <= x 
ghci> isOverlappingIntervals (2, 3) (0, 1) 
True
```

riz0id: Thanks, I have fixed this.

tools/canonicalize-proto-file/Main.hs
```diff
@@ -11,28 +10,110 @@
...
 instance ParseRecord (Args Wrapped)
+
 deriving instance Show (Args Unwrapped)
 
+-- | Check if the given values is some given bounds (inclusive).
+between :: (Ord a, Enum a) => a -> (a, a) -> Bool
+between x i = fst i <= x && x <= snd i
+
+-- | @('isOverlappingIntervals' a b :: 'Bool')@ is a relation between two 
+-- interval-like types @a@ and @b@ that is 'True' when the two intervals are
```
j6carey, line 75: I think the word `types` is confusing here because `a` and `b` are terms, not types.

tools/canonicalize-proto-file/Main.hs
```diff
@@ -11,28 +10,110 @@
...
+normalizeIntervals :: (Ord a, Enum a) => [(a, a)] -> [(a, a)]
+normalizeIntervals = mergeIntervals . filter \(x, y) -> x <= y
+
+-- | Returns a sorted list that contains all intervals from the minimal set of 
+-- intervals to represent the given list of intervals. 
+--
+-- "Merges" overlapping intervals in a list of intervals. Think disjunctive 
+-- normal form.
+mergeIntervals :: (Ord a, Enum a) => [(a, a)] -> [(a, a)]
+mergeIntervals = foldr step [] . sort
```
j6carey, line 109: This implementation looks correct, but could you provide a unit test of some sort?  For example, you could restrict to, say, [0 .. 99], then exhaustively iterate through the two-digit numbers, checking that each number that is in none of the input intervals is in none of the result intervals, and that each number that is in at least one of the input intervals is in exactly one of the output intervals.  That would avoid creating a reference implementation as such--it would just check the desired properties, and with only 100 numbers to check, it should finish promptly.

riz0id: Yeah, I can do. I'll move the code for merging intervals into `Proto3.Suite.DotProto.Internal` and import it into the `canonicalize-proto-file/Main.hs` file. This way both the executable and the test suite have access to the code.

nix/overlays/haskell-packages.nix
```diff
@@ -275,13 +275,27 @@ in {
...
+                  libraryHaskellDepends = [
+                    base bytestring containers tagged template-haskell text time
+                  ];
+                  testHaskellDepends = [
+                    base bytestring containers HUnit tagged text time transformers
+                  ];
+                  description = "Convert values from one type into another";
+                  license = lib.licenses.mit;
+                }) {}
+              else haskellPackagesOld.witch;
```
j6carey, line 298: Did this get added in by mistake, as a result of a merge conflict?

tests/Test/Proto/Interval.hs
```diff
@@ -0,0 +1,93 @@
...
+
+--------------------------------------------------------------------------------
+
+overlappingIntervals :: (Bounded a, Integral a) => Gen ((a, a), (a, a))
+overlappingIntervals = do 
+  x <- Gen.integral Range.constantBounded
+  y <- Gen.integral Range.constantBounded
+
+  u <- Gen.integral (Range.constant x y) 
+  v <- Gen.integral (Range.constant y maxBound)
```
j6carey, line 27: Shouldn't you sometimes put `(u, v)` before `(x, y)`, not just after?

tests/Test/Proto/Interval.hs
```diff
@@ -0,0 +1,93 @@
...
+        -- i2  = (5, 10)
+        -- i1' = (0, 5) 
+        -- i2' = (5, 10) 
+        -- @
+        --
+        -- In all other cases we are gauranteed to have disjoint @i1'@ and 
+        -- @i2'@, thus @'not' . 'isOverlappingIntervals'@ should be implied.
+        unless (x == y || u == v || x == v || y == u) do
+          let i1' :: (Int8, Int8)
+              i1' = (min (fst i1) (fst i2), min (snd i1) (snd i2))
```
j6carey, line 60: Don't the `fst` and `snd` components already have names you can use?

j6carey (changes requested):
Can you test `normalizeIntervals` itself, not just `joinIntervals`?

tests/Test/Proto/Interval.hs
j6carey: Thanks.

j6carey (approved):
Thanks!

PR #287, opened 2025-08-12
Remove `ci.yml`

riz0id:
Removing `ci.yml` to see if it gets rid of the lingering GHC 8.10.7 awaiting status issue in CI since it has persisted even after I have disabled workflows in this repository.


riz0id:
The issue was in branch settings.

PR #289, opened 2025-09-11
v0.9.4: Support optional fields

j6carey:
Older versions of the proto3 specification did not include
optional fields, but the current version recommends them.
In this change we add support for optional fields
(but in the process require proto3-wire >= 1.4.6).

This change also renames and restructures the cardinality
descriptions in the experimental `Proto3.Suite.Form` modules
to better reflect protobuf terminology and encoding.  Note
that in particular, encoding of `oneof` fields now expects
a `Maybe` or `Identity` type to reflect potential omission,
which is a breaking change to experimental usage.

And the parser now rejects attempts to put `optional`,
`repeated`, or `map` fields within a oneof.  (Technically
the old parser also rejected `optional` within `oneof`
because it rejected `optional` in general.)

### DIFF
--- a/src/Proto3/Suite/Form/Encode/Core.hs
+++ b/src/Proto3/Suite/Form/Encode/Core.hs
@@ -601,7 +601,10 @@ instantiatePackableField protoType elementType encoder packedEncoder promotions 
 
       instance FieldForm 'Optional $protoType (Identity $elementType)
         where
-          fieldForm _ _ !fn (Identity x) = $encoder fn x
+          fieldForm _ _ = coerce
+            @(FieldNumber -> $elementType -> Encode.MessageBuilder)
+            @(FieldNumber -> Identity $elementType -> Encode.MessageBuilder)
+            $encoder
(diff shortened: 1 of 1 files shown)


src/Proto3/Suite/Form/Encode.hs
```diff
@@ -408,7 +408,10 @@ instance ProtoEnum e =>
 
 instance FieldForm 'Optional 'Bytes (Identity RB.BuildR)
   where
-    fieldForm _ _ !fn (Identity x) = Encode.bytes fn x
+    fieldForm _ _ = coerce
+      @(FieldNumber -> RB.BuildR -> Encode.MessageBuilder)
+      @(FieldNumber -> Identity RB.BuildR -> Encode.MessageBuilder)
```
riz0id: I don't know if you've ever used this before, but you can use wild cards in type applications and GHC will fill the hole with inference. So you could shorten this to `@(_ -> _ RB.BuildR -> _)` I think, if you wanted.

PR #307, opened 2026-05-14
Fix bug where `--extraInstances` ignored standalone instance declarations

riz0id:
`InstD` was the type of declaration that was pick up by the `--extraInstances` mechanism, but `DerivD` is the constructor for standalone deriving declarations.  This causes standalone deriving instance declarations to be ignored when splicing in extra instances with the `--extraInstances` flag.


tests/Test/Proto/Generate/CodeGen.hs
```diff
@@ -0,0 +1,27 @@
...
+extraInstanceParsing :: TestTree
+extraInstanceParsing =
+  testCase "getExtraInstances includes standalone deriving declarations" $ do
+    logger <- initLogger
+    result <- runExceptT $ getExtraInstances logger "test-files/extra_instances_deriving.hs"
+    case result of
+      Left err -> assertFailure (show err)
+      Right (_imports, decls) ->
+        assertEqual "expected 2 declarations (1 instance + 1 standalone deriving)"
+          2 (length decls)
```
j6carey: Should this test also check that the class names are correct?

proto3-suite.cabal
```diff
@@ -29,6 +29,7 @@ data-files:
 extra-source-files:
   CHANGELOG.md,
   gen/.gitignore
+  tests/Test/Proto/Generate/CodeGen.hs
```
j6carey: Why is the new file listed here?

tests/Main.hs
```diff
@@ -61,6 +61,7 @@ import           TestDhall
 
 import qualified Test.Proto.Generate.Name
 import qualified Test.Proto.Parse.Option
+import qualified Test.Proto.Generate.CodeGen
```
j6carey: There is also a module called `TestCodeGen`.  I was initially confused between the two modules; perhaps others would be as well.  Could there be some kind of different naming to distinguish the two?

riz0id: Yeah, I realize. I can move it into `TestCodeGen` to avoid confusion, but also slowly migrating away from the `turtle` tests might be a good thing even if its a little confusing to contributors for some time.

proto3-suite.cabal
riz0id: I meant to list the `test-file` and copied the wrong relative path from my editor.

tests/Test/Proto/Generate/CodeGen.hs
riz0id: Yes

tests/Main.hs
j6carey: Or find something that distinguishes the two blocks of code and put that in the name?

riz0id: I'll fold it into `TestCodeGen` I don't know if I feel confident about the choice on second thought.

tests/Test/Proto/Generate/CodeGen.hs
j6carey: Thanks.

proto3-suite.cabal
j6carey: OK, thanks.

tests/Main.hs
j6carey: What you have is good enough to merge; if you want to split it out later that's fine.

j6carey (approved):
Thanks!

PR #309, opened 2026-05-18
Remove `Anonymous` identifier constructor

riz0id:
The `Anonymous` constructor has always been redundant. This PR refactors identifiers so that `Anonymous` can be removed. This simplifies code generation and renaming cases.


riz0id:
Closed https://github.com/awakesecurity/proto3-suite/pull/294 in favor of this PR. This is the minimal change necessary to remove `Anonymous`.

riz0id:
This enables: https://github.com/awakesecurity/proto3-suite/pull/310

j6carey (approved):
In future we should work toward eliminating the `Maybe` from the type of `dotProtoFieldName`, but this is an improvement in that direction.

